PR #17, opened 2026-06-08
fix(codegen): break generated server.ts cycle + return-type inference & DX

prisis:
## Why

`pnpm dev` in any Cirrus app crashed at startup with a file-less, undebuggable error:

```
TypeError: Cannot read properties of undefined (reading 'string')
    at runInRunnerObject (workers/runner-worker/index.js:107:3)
    at getWorkerEntryExportTypes (workers/runner-worker/index.js:246:24)
```

Root cause was **our codegen**, not the Cloudflare plugin: the generated `_generated/server.ts` both re-exported the builders user code imports (`v`/`query`/`mutation`) **and** imported every user function module to build `CIRRUS_FUNCTIONS`. That's a circular import — the user-module imports hoist above `export const v = …`, so a module's top-level `mutation({ args: { x: v.string() } })` ran while `v` was still `undefined`. Under Vite's SSR module-runner the read returns `undefined` (not a TDZ error), and workerd's RPC boundary dropped the real user frame.

## What's in here

**Core fix** — `fix(codegen): split generated server.ts to break cycle`
- `server.ts` is now builder-only (no user-module imports → no cycle).
- New `_generated/functions.ts` holds `CIRRUS_FUNCTIONS` / `dispatchCirrusFunction` / `createCaller` / `CIRRUS_MIGRATIONS`, importing the user modules + a type-only edge back to `server.ts`.
- `shard.ts` and worker entries import the registry from `./functions.js`.

**DX** — `refactor(vite)` + the hint plugin
- `@cirrus/vite` wraps the Cloudflare plugin's startup hooks and appends an actionable hint (circular import / stale codegen) to the otherwise file-less runner-worker error — written to `.stack`, which Vite's CLI prints.

**Templates** — were already broken (`main` pointed at a generated file with no `fetch`/`ShardDO` export). Each now has a real `src/server.ts` Worker entry, `main` repointed, and `schema.ts` switched to the default export codegen requires.

**Return-type inference** — `feat(codegen): infer return types from non-exported local types`
- A handler annotated `: Promise<PostDoc[]>` where `PostDoc` is a local, non-exported interface used to erase to `unknown` (so `useQuery` returned `unknown` and forced casts). Codegen now **structurally expands** the type to its real shape — `{ _id: Id<"posts">; …; imageKey?: string }[]` — keeping `Id`/`Doc`/primitives verbatim and rendering optionals as `name?: T`. It **bails back to `unknown`** for anything it can't faithfully reproduce, so it is never worse than before, only more precise. Also closes a latent `TS2304` hole for anonymous objects embedding an unreachable interface.

**Thermo review follow-ups** (a [/thermos](#) double audit ran on the branch):
- `fix(studio)`: a copy-pasteable README snippet imported `CIRRUS_FUNCTIONS` from the now-wrong path → fixed (it silently broke studio function discovery).
- `fix(runtime)`: widened `FunctionRegistryLike` to accept the generated registry directly (dropping `as unknown as FunctionRegistryLike` from all four worker entries) and made the documented "queries/mutations/actions only" discovery contract real (omit `stream`).
- `refactor(codegen)`: extracted shared emitter helpers (byte-identical output); `refactor(vite)`: simplified the hook wrapper.

## Verification

- Dev server boots clean (`ready in ~2s`); the hint renders in real dev output (confirmed with an injected throw).
- All examples + playground typecheck with the richer generated types.
- Tests: codegen 120 ✅, runtime 243 ✅, vite 53 ✅. Every consumer package (`react`/`client`/`studio`/`do`/`cli`) typechecks.
- The emitter refactor + example regeneration produce zero unexpected drift (snapshot test green).

## Follow-ups (not in this PR)
- Convex-style `returns:` validator for runtime return *validation* (types are already precise).
- Optional codegen warning on the rare remaining `unknown` fallback.


prisis:
Superseded — the changes were committed directly onto `alpha` instead of merging via this branch. Closing; the 7 commits now live on local `alpha`.

PR #30, opened 2026-06-26
feat(queue): add queues, pipelines, secrets bindings + studio queues page

prisis:
## What

Adds three Cloudflare-platform bindings the framework was missing, plus a Studio observability page for queues. Closes the most conspicuous gaps from a platform-coverage audit (Queues was entirely absent; Pipelines was a dangling emitted type; Secrets Store had no surface).

Net **+1 package** — Pipelines folded into an existing package and Secrets shipped as a core built-in rather than two more standalone installs.

### `@lunora/queue` (new package)
- `defineQueue` in `lunora/queues.ts` → typed `ctx.queues.<name>` producers on **Mutation/Action** ctx (enqueue is a side effect, excluded from `QueryCtx`).
- Generated worker `queue()` **push** consumer (handler decoupled from `@lunora/server`, dispatches via `ctx.run` → `/_lunora/scheduler/dispatch` like workflows), or `mode: "pull"` for an HTTP pull consumer.
- Codegen emits `_generated/queues.ts` (push registry) + `LUNORA_QUEUES` specs; `@lunora/config` reconciles `queues.producers[]` / `queues.consumers[]` and shape-validates them.
- `vis generate lunora-queue` scaffolder.

### Pipelines → folded into `@lunora/analytics`
- `createPipelines` → `ctx.pipelines` (ActionCtx-only, R2-backed streaming ingestion). The previously-dangling emitted `PipelineClient` type now resolves; `ctx.pipelines`-only feature detection so a plain analytics import doesn't wire it.

### Secrets Store → core `ctx.secrets` in `@lunora/server`
- `ctx.secrets.get(name)` on every ctx. The structural binding check rejects KV/R2/DO/Queue/Analytics bindings (they also expose `.get`) so a wrong name throws a directed error instead of silently reading the wrong binding.

### Studio Queues page
- Read-only table of declared queues (export, name, mode, binding, dead-letter), gated by `studioFeatures.queues`, fed by the new `__lunora_admin__:listQueues` DO admin RPC + `queuesMetadata()` hook. Wire contract drift-guarded across `@lunora/do` and `@lunora/studio`.

## Verification
- Build **41/41**; tests: server 313, do 837, codegen 365, config 360, runtime 395, queue 15 — all green; lint + types clean.
- End-to-end smoke: codegen → `_generated/queues.ts` + `ctx.queues`/`ctx.secrets`, infer detects push/pull, reconcile writes correct wrangler config.
- Reviewed via **thermos** (two passes): no critical/high bugs. Addressed the secrets `.get` collision (real footgun) and added the queue-metadata drift guard.

## Notes
- Dispatch runner in `queue/src/run-context.ts` is a deliberate parallel of `@lunora/workflow`'s (siblings, no shared `@lunora` dep — the framework's decoupling convention). Consolidating the runner + `FunctionReference` triple across queue/workflow/scheduler is a worthwhile follow-up but a cross-package refactor of shipped code, intentionally left out of this change (documented in-code).


prisis:
## 🌡️ Thermos review — queues / pipelines / secrets

Two parallel passes (a diff-scoped **bug/security branch-audit** + a **code-quality** review) over `feat/cloudflare-queues-pipelines-secrets` vs `alpha`, then synthesis. All findings below are **fixed and pushed** on this branch.

### Verdict
**Clean — no Critical / High / Medium.** An unusually disciplined diff: null-prototype producer map, `assertIdentifier` + `JSON.stringify` on every interpolated identifier/literal, bound-parameter SQL, fail-open studio gating, and graceful directed-error stubs for every missing binding. The bug/security pass traced the full producer/consumer wiring (`env → worker.queue → dispatchQueueBatch → registry keyed by batch.queue === queues.consumers[].queue`), the secrets-store reader, the pipelines surface, all codegen emission, config reconcile/validation, the DO admin path, and the studio panel, and found nothing exploitable.

### Fixes applied

| # | Sev | Area | Fix |
|---|-----|------|-----|
| 1 | Med | `@lunora/server` `create-secrets` | The "no Secrets Store binding named X" error pointed users at `.dev.vars`, but `ctx.secrets` reads a Secrets Store binding (`env[name].get()`) — a `.dev.vars` value is a plain string that can never satisfy the check. Error now points at the `secrets_store_secrets[]` binding and says so. |
| 2 | Low | `@lunora/server` `create-secrets` | A Cloudflare **Workflows** binding exposes `get(id)` alongside `create`/`createBatch`, so it structurally passed the `{ get }` secret check. Added `create`/`createBatch` to the non-secret disambiguators. |
| 3 | Low | `@lunora/queue` `create-queues` | Built the `ctx.queues` producer map with `Object.create(null)` + `Object.hasOwn`, so a queue named `constructor` / `toString` / `hasOwnProperty` resolves to the directed "no queue named X" error instead of an inherited Object member. |
| 4 | Low | `@lunora/queue` `define-queue` | Documented that a push handler's `ctx.run(...)` dispatches over the **admin-authenticated** path (system identity, **end-user RLS not applied**) — handlers must validate `message.body` before forwarding it into a privileged mutation. |
| 5 | Low | `@lunora/config` `infer-bindings` | The `usesPipelines` capability matched `from "@lunora/pipelines"` — a package that doesn't exist (Pipelines ships from `@lunora/analytics`, consumed via `ctx.pipelines`). The "add a `pipelines` binding" wrangler hint never fired. Now keyed off the `ctx.pipelines` access (an analytics-only import does **not** flip it); both hint strings + the test that encoded the dead import updated. |
| 6 | Low | `@lunora/codegen` `discover-queues` | An empty `name: ""` override flowed into the registry key + reconciled wrangler queue name. Mirrored the runtime `defineQueue` guard with a located diagnostic at discovery; added a `discover-queues` regression test. |

### Verification
`@lunora/{queue,server,config,codegen}` — `lint:types` + `eslint` clean; queue (15), server create-secrets (5), config infer/reconcile (65), codegen discover-queues (4) + emit (15) suites green.

PR #32, opened 2026-06-27
feat: extending db 

prisis:
Close the four feature gaps versus better-drizzle on the typed `ctx.db`
table delegates (`ctx.db.<table>.*`):

- `upsert` / `upsertMany` — insert-or-update keyed by a unique `target`
  (one field or a tuple). Returns `{ id, created }`.
- `exists(where?)` — boolean existence check (reuses `findFirst`, no
  count scan).
- `select` — `findMany`/`findFirst` field projection. Trims the returned
  payload to the named columns plus `_id`/`_creationTime`; applied after
  relation resolution + cursor encoding so paging and dependency tracking
  are unaffected. Engine-level in `@lunora/do` and `@lunora/sql-store`.
- `insert({ skipDuplicates })` — a UNIQUE breach resolves to `null`
  instead of throwing.

upsert/upsertMany/exists/skipDuplicates are implemented as compositions
of `findFirst`/`insert`/`patch` at the facade layer (`bindTableFacade`),
so they inherit RLS, secure-by-default, and `.global()`/D1 routing for
free — the RLS middleware re-binds the same facade over its wrapped
writer, gating each composed step. No new low-level writer methods or
RLS-middleware changes were needed. The facade detects a UNIQUE conflict
structurally (ConflictError `code`/`kind`) to avoid a runtime dependency
on `@lunora/do`.

Tests: facade unit tests (mocked writer), an end-to-end integration test
over the real DO writer + a real SQLite UNIQUE index, and engine-level
`select` projection tests.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01UxkuKVSsfq6g72A3cbDxpJ

## Checklist

- [ ] Commit messages follow the [Conventional Commits](./commit-convention.md) style (`feat(scope): subject`)
- [ ] Added or updated tests covering the change
- [ ] Updated relevant docs in `apps/docs` (or `README.md` for package-local docs)
- [ ] No `package.json` files in `packages/*` modified outside the touched package
- [ ] If a new package was added: `project.json` has `type:package` and a `category:*` tag
- [ ] If migration impact: noted upgrade steps in the PR description and changelog entry

### DIFF
--- a/packages/codegen/__tests__/discover-schema.test.ts
+++ b/packages/codegen/__tests__/discover-schema.test.ts
@@ -43,6 +43,32 @@ describe("discoverSchema", () => {
         expect(schema.tables.find((table) => table.name === "messages")?.externallyManaged).toBe(false);
     });
 
+    it("captures `.softDelete()`, injecting the marker column so Doc carries it", () => {
+        expect.assertions(4);
+
+        const { project, schemaPath } = projectWith(`
+            import { defineSchema, defineTable, v } from "@lunora/server";
+
+            export const schema = defineSchema({
+                posts: defineTable({ title: v.string() }).softDelete(),
+                logs: defineTable({ line: v.string() }).softDelete({ field: "removedAt" }),
+            });
+        `);
+
+        const schema = discoverSchema(project, schemaPath);
+        const posts = schema.tables.find((table) => table.name === "posts");
+        const logs = schema.tables.find((table) => table.name === "logs");
+
+        expect(posts?.softDelete).toStrictEqual({ field: "deletedAt" });
+        expect(logs?.softDelete).toStrictEqual({ field: "removedAt" });
+
+        // The injected column flows into the emitted Doc as an optional number.
+        const dataModel = emitDataModel(schema);
+
+        expect(dataModel).toContain("deletedAt?: number;");
+        expect(dataModel).toContain("removedAt?: number;");
+    });
+
     it("captures searchIndex name + field + filterFields", () => {
         expect.assertions(3);
 
--- a/packages/codegen/src/discover-schema.ts
+++ b/packages/codegen/src/discover-schema.ts
@@ -315,8 +315,26 @@ interface TableBuilderAccumulator {
     relations: RelationIR[];
     searchIndexes: SearchIndexIR[];
     shardMode: TableIR["shardMode"];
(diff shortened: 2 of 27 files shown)


prisis:
## 🌡️ /thermos:thermos review — findings & fixes

Two parallel passes (branch bug/security audit + code-quality) over the soft-delete / Drizzle-features diff. Summary of what was found and what landed on this branch.

### MEDIUM — `rank()` / `rankPage()` leaked soft-deleted rows ✅ fixed
The rank companion is a precomputed shadow table read **verbatim** at read time — unlike `search` / `aggregate` / `count`, which re-scan the live table and apply the soft-delete scope. So a soft-deleted row kept its rank position and still surfaced in rank pages, disagreeing with every list read that hides it. The same gap let a Vectorize vector survive a soft-delete. The off-index marker column also let the `rankIndexFieldsUnchanged` fast path skip the companion delete on soft-delete (and the re-insert on restore).

**Fix** (`cb69dec8`): a soft-delete now *removes* the row from the unscopeable companions — the rank companion entry is dropped (`syncRanks(..., undefined)`) and `onWrite` fires `op: "delete"` so the Vectorize hook removes the vector; `restore()` re-adds both, forcing a pure `INSERT` to defeat the rank fast-path (guarded to the deleted→live transition only). Both engines (`@lunora/do` + `@lunora/sql-store`). Regression tests cover rank/Vectorize exclusion + restore + `hardDelete`, and that the soft-delete scope AND-composes with an injected RLS `baseWhere` (neither bypasses the other).

### LOW — docs inaccuracy ✅ fixed
The `softDelete()` JSDoc named a **phantom `withDeleted`** option alongside `includeDeleted` — only `includeDeleted: true` exists (`f10b9d4c`). Clarified that `includeDeleted` is a *read scope, not access control* (anyone who can run the read can set it), and that a unique index still rejects a row colliding with a soft-deleted one (the row physically persists).

### Verification
- `@lunora/do`, `@lunora/sql-store`, `@lunora/server` — `tsc --noEmit` clean.
- Full `@lunora/do` suite green (the one transient `ctx-db.vectors` failure was an unbuilt `@lunora/vectors` dist, not a regression).
- pre-commit prettier + eslint clean on all touched files.

PR #33, opened 2026-06-27
perf(runtime): skip route lookup when no custom routes

prisis:
Guard the `${method} ${pathname}` string allocation and the two route-map
reads behind an `options.routes` presence check, so the common composed-worker
path (no custom routes) avoids that per-request work on the hot dispatch path.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_012e9XTD4pVzwsvxd8RP3YW6

## Checklist

- [ ] Commit messages follow the [Conventional Commits](./commit-convention.md) style (`feat(scope): subject`)
- [ ] Added or updated tests covering the change
- [ ] Updated relevant docs in `apps/docs` (or `README.md` for package-local docs)
- [ ] No `package.json` files in `packages/*` modified outside the touched package
- [ ] If a new package was added: `project.json` has `type:package` and a `category:*` tag
- [ ] If migration impact: noted upgrade steps in the PR description and changelog entry


prisis:
## 🌡️ Thermos review — findings & fixes

Ran the thermo-nuclear branch audit (bugs/security) + code-quality review on this branch. Two findings, both fixed in this push.

### 🔴 HIGH (security) — AOT validator bypass on optional refinements
`packages/codegen/src/compile-validator.ts` — `v.optional(v.string()).check(...)` lowers to an `optional` IR node carrying `hasRefinement: true`. The optional-field compile branch recursed straight into `inner`, compiling a bare guard and **silently dropping the `.check()` predicate** — the AOT hot path then accepted input the interpreted parser rejects (a validation bypass). Fixed by consulting the optional wrapper's own flags (`hasRefinement` / `sourceText` / column modifier) and declining to compile, exactly as `compileNode` does for every other node. Regression tests added (declines optional `.check()` on string + number; plain optional still compiles).
→ commit `80eca20b`

### 🟡 MEDIUM (quality) — route fast-path was a no-op
`packages/runtime/src/create-worker.ts` — the per-request `if (options.routes)` guard was meant to skip the custom-route lookup on the common composed-worker path, but the generated workers always pass a literal `routes: {}`, so the guard was truthy on every request and never skipped the `"METHOD path"` allocation + object reads. Now resolved once at construction (`customRoutes`, empty object treated as none). Restored the named `methodAndPath` local. Tests added for the no-routes 404 path and that an empty `routes: {}` does not shadow a matching httpRouter route.
→ commit `42baa770`

All runtime (53) + codegen compile-validator (19) tests pass; `lint:types` clean for both packages.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #34, opened 2026-06-27
perf(codegen): ahead-of-time compile argument validators

prisis:
## Summary

Speeds up per-request function dispatch by compiling argument validators **ahead of time** at `lunora codegen`, inspired by Elysia's JIT/Sucrose and modelled on [`zod-compiler`](https://github.com/gajus/zod-compiler). Because Cloudflare Workers forbid runtime `eval`/`new Function`, the specialised validator is emitted as static source at build time (no runtime codegen) rather than JIT-compiled in-process.

Touches **`@lunora/codegen`** (the compiler + emit wiring), **`@lunora/values`** (the install seam), and **`@lunora/runtime`** (a small dispatch micro-opt).

How it works:

- `@lunora/codegen` emits a specialised, zero-allocation parser per eligible function from the existing `ValidatorIR`, and installs it onto the function's live `.args` object — the same reference the procedure builder validates against.
- `@lunora/values` `parseValidatorMap` consults the installed parser as a fast path (WeakMap-keyed). The compiled code is built to be **sound, not complete**: any check it's unsure about returns a `DEFER` sentinel, so **every error message and every unmodelled validator** (`.check` refinements, `v.union`, `v.record`, `v.from`, referenced-validator identifiers) still flows through the single interpreted implementation. Error contracts can't drift.
- Refinements are captured structurally on the IR (`ValidatorIR.hasRefinement`) and declined by the compiler, so a `.check(...)` predicate is never silently skipped.

Measured locally (`vitest bench`, valid input): **~2.1× / ~8.2× / ~13.9×** faster than the interpreted parser for scalar args / object+array / array-of-20-objects respectively — the win grows with schema size/nesting, where validation cost actually lands.

Also includes a tiny `@lunora/runtime` dispatch micro-opt: skip the custom-route map lookup (and its per-request string allocation) when no `routes` are configured — the common composed-worker case.

## Test plan

- [x] `pnpm lint:types` passes (`@lunora/values`, `@lunora/codegen`, `@lunora/runtime`, `@lunora/server`)
- [x] `pnpm lint:eslint` passes on all changed files (pre-commit `eslint --fix` clean on every commit)
- [x] Affected suites pass: `@lunora/codegen` (381), `@lunora/values` + `@lunora/server` (328) — includes a **differential test** comparing compiled output against the interpreted oracle across a broad input corpus, and an end-to-end builder-dispatch test
- [x] Codegen regenerated cleanly — the committed `_generated/functions.ts` golden fixture is byte-stable
- [ ] workerd-based Vitest pool tests — not run (no Durable Object code changed; runtime change is the worker fetch path, covered by `@lunora/runtime` unit tests)
- [ ] Manual smoke in a playground app — n/a (no public API or behaviour change; opt-in fast path with interpreted fallback)

## Checklist

- [x] Commit messages follow Conventional Commits (`perf`/`fix`/`refactor`/`test` with package scopes)
- [x] Added/updated tests covering the change (differential parity, refinement/reference declines, seam install + fallback, builder e2e, benchmark)
- [ ] Updated docs in `apps/docs` — not needed (internal optimization, no API surface change)
- [x] No `package.json` modified outside the touched packages (none modified)
- [x] No new package added
- [ ] No migration impact

## Notes for reviewers

- The soundness contract is the load-bearing invariant — start at `packages/codegen/src/compile-validator.ts` (the doc comment explains why the fast path only needs to be *sound*, never *complete*) and `packages/values/src/validator-map.ts` (the `DEFER` seam). The differential test (`packages/codegen/__tests__/compile-validator.test.ts`) is the parity guarantee.
- A self-review caught and fixed one real soundness hole during development (`fix(codegen): decline AOT compile of unresolved validator refs`): a validator referenced by identifier lowers to `{ kind: "any", sourceText }`, which must decline rather than compile to a guard-free pass-through. There's a regression test for it.
- **Deferred (not in this PR):** per-function context tree-shaking. Investigated and deliberately left out — the safe lazy-getter design is blocked by the supported `{ ...ctx }` middleware-spread pattern, and the existing project-level + `isAction` gating already captures the safe wins. Happy to follow up if wanted.

---
_Generated by [Claude Code](https://claude.ai/code/session_012e9XTD4pVzwsvxd8RP3YW6)_


prisis:
Closing as superseded by #33, which merged the same work (AOT-compiled argument validators + the runtime route-lookup micro-opt) into `alpha`. This PR was opened from a pre-#33 base and duplicates it — alpha also carries follow-up improvements not in this branch (the optional-field `.check` decline in `compileField`, and the unrelated `softDelete` addition). Nothing here to merge.

---

PR #45, opened 2026-06-29
feat(hono): add Hono single-worker composition adapter

Shakenokirimi12:
## What

Adds **`@lunora/hono`** — the Hono framework integration. Hono has no reactive UI layer, so this package owns the **server-composition seam only** (reactivity stays in `@lunora/client`), mirroring `@lunora/astro` / `@lunora/nuxt`.

## API

- **`mountLunora(new Hono())`** — mounts Lunora's realtime plane (`/_lunora/*` RPC + WebSocket + admin) as a Hono middleware and returns the **same typed app**, so it slots straight into `const app = …` and everything after stays 100% plain Hono. `shardDO` defaults to the conventional `env.SHARD` binding, so the common case needs no options.
- **`lunora(options?)`** — the bare middleware, for a custom mount path / `app.use`.
- **`withLunora`** — re-export of the shared `withFrameworkWorker` (the same composer behind `@lunora/svelte/worker`, `@lunora/vue/worker`, `@lunora/astro`) for the inverse "Lunora owns the entry" direction.
- **`@lunora/hono/server`** — re-exports the framework-neutral `@lunora/client/ssr` reactive-loader helpers.

```ts
const app = mountLunora(new Hono<{ Bindings: Env }>());
app.get("/", (c) => c.text("hi"));
export default app;
```

Optional `hono` peer dependency. No new runtime logic — the composition reuses `@lunora/runtime`'s `createWorker` / `withFrameworkWorker`.

## Tests

`__tests__/index.test.ts` — 9 passing: default `env.SHARD` RPC routing, non-`/_lunora` fall-through to Hono, `/_lunora/ws` upgrade routed to Lunora (not the Hono fall-through), explicit `shardDO` override, `(env) => options` factory, clear error when no shard namespace resolves, hand-mounted middleware, and `withLunora` inverse composition.

## Checks (local)

- `lint:eslint` ✅  `lint:types` ✅  `lint:prettier` ✅ (source)  `test` ✅ 9 passed  `build` / `build:prod` ✅

Also adds a `@lunora/hono` row to the package table in `AGENTS.md`.

## Summary by CodeRabbit

* **New Features**
  * Added a new Hono integration package for embedding Lunora into existing apps, including middleware mounting and server-side helpers.
  * Exposed server-side data loading and hydration utilities for smoother SSR workflows.
* **Documentation**
  * Added package docs with setup, usage patterns, custom path mounting, and runtime proxy guidance.
* **Chores**
  * Added package metadata and updated build, lint, test, TypeScript, Prettier, and release configuration.
  * Included the package license text and publish settings.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the project's license and that you can use, modify, copy, and redistribute this contribution under those terms.


prisis:
First off — thank you for this! 🙏 The PR was well-built: clean composition, the `(env) => options` factory, the guarded `executionCtx`, the WS-upgrade pass-through, and a thorough 9-case test suite. It also surfaced exactly the right question, and digging into it changed how we want to support this across the board.

**The decision: we're generalizing the seam instead of shipping a per-framework package.**

Working through your PR, we realized the whole integration is framework-neutral — `createWorker(...).fetch(request, env, ctx)` already speaks the Web Fetch contract, so the only thing an adapter does is pluck `(request, env, ctx)` out of the host's context. The pieces your PR added on top (the `env.SHARD` default + factory in `resolveOptions`, the no-op `ExecutionContext`, returning the `101` upgrade `Response` verbatim) aren't Hono-specific at all — and they were already being re-implemented separately inside `@lunora/nuxt`.

So we lifted that seam into `@lunora/runtime` as a single shared helper, **`createLunoraHandler(options?)`** → `(request, env, ctx?) => Response`. Every web-standard framework that runs on Workers (Hono, Nitro/h3, Elysia, any WinterCG host) becomes a 1–2 line mount over it:

```ts
import { createLunoraHandler } from "lunorash/runtime";

const lunora = createLunoraHandler();
app.use("/_lunora/*", (c) => lunora(c.req.raw, c.env, c.executionCtx)); // Hono
```

It keeps everything your package did — `shardDO` defaults to `env.SHARD`, the `(env) => options` factory form, the WS pass-through — just without a package per framework. We're reserving dedicated packages for frameworks that need real build/module wiring beyond a fetch bridge (e.g. **Nuxt**, which registers a server handler, the `#lunora/app` virtual, and re-exports `ShardDO`).

Landed on `alpha` in `5473874d`, with docs (incl. Nitro/Elysia examples + the host contract) in the [Bring your framework](https://github.com/anolilab/lunora/blob/alpha/apps/docs/src/content/docs/frameworks/bring-your-framework.mdx#mount-lunora-inside-your-framework) guide. Your PR is what got us there — closing this in favor of that shared approach. Thank you again, and please keep the contributions coming! 🚀

PR #51, opened 2026-06-29
fix(container): inline patched @cloudflare/containers

prisis:
## Summary

Moves `@cloudflare/containers` from a runtime **dependency** to a **devDependency** of `@lunora/container`, and lets packem inline its source into our `dist`. This is what makes the fixes we carry as a `pnpm patch` actually reach published consumers — a runtime dep would resolve to the pristine, still-buggy `0.3.7` on a user's machine; bundling the patched source bakes the fixes into what we ship.

## The backported fixes (`patches/@cloudflare__containers@0.3.7.patch`)

- **[cloudflare/containers#191](https://github.com/cloudflare/containers/pull/191)** — alarm-scheduler spinloop under concurrent callers. The base class paired an in-memory `setTimeout` sleep with an unconditional `setAlarm(Date.now())`, so any concurrent `scheduleNextAlarm()` spun the alarm at sub-second cadence, saturating the DO event loop and canceling queued WebSocket upgrades. Fix: re-arm a single storage alarm at the next deadline and make `scheduleNextAlarm()` idempotent (no-op if an earlier alarm is already set). **Production bug** we inherit by extending `Container`.
- **[cloudflare/containers#230](https://github.com/cloudflare/containers/pull/230)** — `NOT_LISTENING_ERROR` constant didn't match the `wrangler dev` runtime message (which omits the leading "the"), so the local readiness loop retried forever. Fix: shorten the constant to a substring of both prod and local-dev messages.

A subclass override was impossible — the buggy logic reads private base-class fields (`this.monitor`/`this.timeout`/`this.resolve`/`this.sleepAfterMs`), so a patch is the only clean route.

## How the inlining works

- packem auto-bundles `devDependencies`; `externals: [/^cloudflare:/]` keeps only the workerd built-in `cloudflare:workers` external.
- The bundled `Container` class + both fixes land in `dist/packem_shared/ContainerProxy-*.mjs`; `do/index.mjs` imports `Container` from it.
- `@lunora/container/do` is now **self-contained** — consumers no longer need `@cloudflare/containers` installed — and the package has **zero runtime dependencies**.

## Verification

- ✅ Shipped bundle's only external import is `cloudflare:workers`; zero runtime/type edges to `@cloudflare/containers` in `dist`
- ✅ Both fixes present in `dist`; `StopParams` / `OutboundHandler` / `ContainerProxy` / `outboundParams` types inlined into the published `.d.ts`
- ✅ 90/90 container tests pass, `tsc --noEmit` clean
- ✅ No other package/app imported `@cloudflare/containers` directly

## Maintenance note

The fixes now live in `dist`, so production builds must run after a patched `pnpm install` (the patch is wired into `patchedDependencies`, so CI does this). When `@cloudflare/containers` is next bumped, pnpm will error if the patch no longer applies — the signal that upstream shipped the fix and both the patch and the inline-vs-external decision can be revisited.

### DIFF
--- a/packages/container/README.md
+++ b/packages/container/README.md
@@ -90,7 +90,9 @@ export const transcode = action.input({ videoId: v.id("videos") }).action(async 
 });
 ```
 
-`ctx.containers` is action-only (container calls are external I/O, like `ctx.fetch`); `.get(name)` handles also expose `start`/`stop`/`destroy`/`getState` lifecycle control plus `renewActivityTimeout()` (keep a busy WebSocket's container awake) and `egress.*` (adjust the allow/deny lists at runtime).
+`.get()` and `.any()` retry the **same** instance through a cold start — when a request lands while Cloudflare is still provisioning (a `503` "no instance", `500` "Failed to start", `429`, or "not listening"), they back off and retry (default 3 attempts) so the provisioning race never reaches your handler. Genuine app `5xx`s pass straight through. Tune or disable per call with `.get(id, { attempts, backoffMs })` (a pre-built `Request` is sent once and not retried, since its body may not be replayable).
+
+`ctx.containers` is action-only (container calls are external I/O, like `ctx.fetch`); `.get(name)` handles also expose `start`/`stop`/`destroy`/`getState` lifecycle control plus `renewActivityTimeout()` and `egress.*` (adjust the allow/deny lists at runtime). HTTP requests and WebSocket frames already keep a busy container awake automatically; `renewActivityTimeout()` is the escape hatch for non-HTTP/non-WS activity.
 
 The config layer (`lunora dev` / `lunora deploy`) reconciles the wrangler `containers[]` entry, the `CONTAINER_*` Durable Object binding, and the SQLite-class migration automatically; `wrangler deploy` builds the Dockerfile with local Docker and pushes it to the Cloudflare Registry.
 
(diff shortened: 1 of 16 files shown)


packages/container/src/do/index.ts
```diff
@@ -115,12 +119,45 @@ class LunoraContainer<Env = unknown> extends Container<Env> {
...
+            await this.resolveSecretsStoreEnv();
+        }
+
+        return super.start(...args);
     }
 
     public override async onActivityExpired(): Promise<void> {
-        // The container slept after its `sleepAfter` idle window. Surfacing it
-        // makes the WebSocket-keepalive gap (cloudflare/containers#147) visible
-        // in the dev log + Studio rather than a silent disappearance.
```
joe345-str, line 123: L121-123 signed off: joe345-str/joeydonner1979@gmail.com

packages/container/src/define-container.ts
```diff
@@ -164,10 +164,31 @@ const assertValidImage = (image: ContainerConfig["image"]): void => {
 };
 
 /**
- * Validate `env`/`buildArgs`/`secrets` naming and reject a name declared in both
- * `env` and `secrets` (where the secret would silently overwrite the static env
- * value at start). All three name sets must be valid env-var names; the
- * collision is rejected at authoring time so the runtime resolver never has to.
```
joe345-str, line 170: L167-170 sign off: joe345-str/joeydonner1979@gmail.com

packages/container/README.md
```diff
@@ -90,7 +90,9 @@ export const transcode = action.input({ videoId: v.id("videos") }).action(async
 });
 ```
 
-`ctx.containers` is action-only (container calls are external I/O, like `ctx.fetch`); `.get(name)` handles also expose `start`/`stop`/`destroy`/`getState` lifecycle control plus `renewActivityTimeout()` (keep a busy WebSocket's container awake) and `egress.*` (adjust the allow/deny lists at runtime).
```
joe345-str, line 93: signed off: joe345-str/joeydonner1979@gmail.com

packages/container/package.json
```diff
@@ -65,10 +65,8 @@
         "fallow:dead-code": "fallow dead-code",
         "fallow:health": "fallow health --score"
     },
-    "dependencies": {
-        "@cloudflare/containers": "catalog:cloudflare"
-    },
```
joe345-str, line 70: L68-70 signed off: joe345-str/joeydonner1979@gmail.com

PR #54, opened 2026-06-29
fix(lfse): wave-4 hardening (redact RPC errors, keyed fan-out, restart-safe diff-emit)

prisis:
Three hardening followups to the local-first sync engine (stacked on `lfse-followups`, the followup branch for #37). Surfaced by a two-pass thermo review; the one finding (066) is fixed.

### `security(do)`: redact unhandled error messages in `errorToResponse`
`ShardDO`'s RPC error path no longer echoes arbitrary `error.message` to clients — it logs the error server-side and returns a static `{ code: "RPC_FAILED", message: "internal error" }`, mirroring `@lunora/runtime`'s `toErrorResponse`. **Reviewed clean** by both passes.

### `perf(client)`: key optimistic fan-out by subscription key
Replaces an O(n) scan over `this.subscriptions.all()` (matching fn/shardKey/argsKey on every optimistic write) with an O(1) `SubscriptionRegistry.key(...)` lookup. Behavior-preserving — the registry already dedups one state per key. **Reviewed clean** by both passes.

### `perf(db)`: cache diff-emit row serialization, restart-safe
`makeDiffEmit` caches each synced row's JSON so an unchanged snapshot re-diffs without re-serializing.

**Fix applied (the one review finding, MEDIUM, confirmed by both passes):** the original split — a caller-owned, restart-persistent `synced` map plus a closure-private JSON cache — desynced when TanStack DB restarts a collection's sync closure. On GC teardown TanStack clears the collection's data (`state.cleanup()` → `syncedData.clear()`) and re-invokes `sync`; the stale `synced` then suppressed the required re-inserts (and churned unchanged rows as spurious updates). The base is now a single internally-owned `key→json` map — fresh per `sync` start, restart-safe by construction, no lockstep invariant. Verified against the TanStack DB 0.6.12 source (`state.js:929 cleanup()`). Adds a sync-restart regression test.

## Verification
- `@lunora/db`: 34 tests pass (incl. new restart regression), typecheck + ESLint clean
- `@lunora/client`: 262 tests pass
- `@lunora/do` + `@lunora/client` + `@lunora/db`: build + typecheck clean, 0 ESLint errors

### DIFF
--- a/packages/db/__tests__/internals.test.ts
+++ b/packages/db/__tests__/internals.test.ts
@@ -108,9 +108,8 @@ const recordingWriter = (): { ops: ({ key: string; type: "delete" } | { type: "i
 
 describe(makeDiffEmit, () => {
     it("emits only the rows that changed between snapshots", () => {
-        const synced = new Map<string, Row>();
         const { ops, writer } = recordingWriter();
-        const emit = makeDiffEmit(synced, writer);
+        const emit = makeDiffEmit(writer);
 
         // First snapshot: two inserts.
         emit(
@@ -127,7 +126,6 @@ describe(makeDiffEmit, () => {
             { type: "insert", value: { _id: "a", text: "1" } },
             { type: "insert", value: { _id: "b", text: "2" } },
         ]);
-        expect(synced.size).toBe(2);
 
         // Second snapshot: `a` changed, `b` removed, `c` added.
         ops.length = 0;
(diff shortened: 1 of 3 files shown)


prisis:
Superseded — the wave-4 hardening (redact RPC errors #70/064, keyed fan-out #59/065, restart-safe diff-emit #60/066) is merged to alpha as granular per-plan PRs.

PR #55, opened 2026-06-29
docs(plans): correct presence phasing in plan 075

prisis:
## What

Corrects a factual misclassification in **plan 075** (auto-elastic fan-out relay tier — PartyKit `partysub` reframed as automatic, invisible transport elasticity). The earlier draft grouped `usePresence` with `whisper` as an opaque, *uniform-by-construction* Phase-2 first adopter.

## Why

Presence is **not** a whisper. `definePresence` ships a `heartbeat` **mutation** + a `listPresent` **reactive query** over a presence table (`packages/server/src/presence.ts` — *"Live queries (subscriptions) drive `listPresent`"*). It therefore flows through the reactive-shape path and the **RLS-uniform gate**, not the opaque whisper channel. Bundling it into Phase 2 would have misrouted an executor.

## Change

- **Phase 2** scoped to `whisper` only (genuinely opaque, single-shard, uniform), with an explicit callout that `usePresence` is excluded and why.
- **Phase 3** names `listPresent` as the canonical RLS-uniform reactive-shape target (typically room-public → passes the gate; highest-fan-out reactive query in practice).
- **Decision §3**, the **Maintenance notes**, and the **README Wave 5** index entry reconciled so all five touch points agree: only `whisper` is uniform-by-construction; every reactive shape — including presence — must pass the gate.

## Scope

Doc-only: `plans/075-*.md` (+38/−13) and `plans/README.md` (+3/−2). No code change.

## Verification

- Surfaced by `/review`; the correction verified against `packages/server/src/presence.ts` and `packages/react/src/use-presence.ts`.
- Two thermos passes (branch audit + code-quality) both returned **clean** — every claim checks out against live code, all touch points internally consistent.

https://claude.ai/code/session_019NXayJPpMpVhqzEd4dLEx4

### DIFF
--- a/plans/075-do-auto-elastic-fanout-relay-tier.md
+++ b/plans/075-do-auto-elastic-fanout-relay-tier.md
@@ -130,8 +130,11 @@ checkpoint)` (≈6111) drains the op range from the owner's SQLite and probes
 3. **RLS-uniform gate signal.** Reuse plan 073's identity-independence
    determination as the _necessary_ condition for promoting a **reactive shape**.
    Confirm: is 073's signal available at the granularity this needs (per shape,
-   at runtime)? If not, what is the minimal extension? whisper/presence are
-   uniform by construction and need no per-shape proof.
+   at runtime)? If not, what is the minimal extension? `whisper` topics are
+   uniform by construction (opaque payload, no per-identity result) and need no
+   per-shape proof; presence (`usePresence`) is **not** uniform-by-construction —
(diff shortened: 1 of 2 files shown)


prisis:
Superseded by #73. The fix targeted `lfse-followups`, which diverged from `alpha` (the wave was squash-merged into alpha instead). #73 carries this same presence-phasing correction onto alpha alongside the plan 075 Phase 1 implementation.

PR #56, opened 2026-06-29
fix: redact RPC error messages + key optimistic fan-out

prisis:
Two hardening followups that are independent of the local-first sync engine and apply cleanly to `alpha`. Cherry-picked from the wave-4 hardening batch; the third commit (066, restart-safe diff-emit) targets `@lunora/db`'s collection layer which doesn't exist on `alpha` yet, so it stays on #54.

### `security(do)`: redact unhandled error messages in `errorToResponse`
`ShardDO`'s catch-all RPC error path no longer echoes arbitrary `error.message` to clients — an unhandled throw may carry SQL fragments, file paths, or internal identifiers. It now logs the raw error server-side and returns a static `{ code: "RPC_FAILED", message: "internal error" }`, mirroring `@lunora/runtime`'s `toErrorResponse`. The typed branches (Conflict / Validation / Lunora errors) are unchanged — those messages are intentional client contract.

### `perf(client)`: key optimistic fan-out by subscription key
Replaces an O(n) scan over `this.subscriptions.all()` (re-matching fn/shardKey/argsKey on every optimistic write) with an O(1) `SubscriptionRegistry.key(...)` + `.get(...)` lookup. Behavior-preserving — the registry already keeps exactly one state per key.

## Verification (on the `alpha` base)
- `@lunora/do` + `@lunora/client`: build + typecheck clean, 0 ESLint errors
- `@lunora/client`: 250 tests pass

### DIFF
--- a/packages/client/__tests__/lunora-client.test.ts
+++ b/packages/client/__tests__/lunora-client.test.ts
@@ -1490,6 +1490,88 @@ describe("lunoraClient", () => {
             expect(received).toEqual([0, 1, 2]);
         });
 
+        it("optimistic update only applies to the subscription matching (fn, args, shardKey) — different args are not mutated", async () => {
+            expect.assertions(2);
+
+            // Two subscriptions to the same function with different args.
+            // A mutation optimistic-targeting args {id:1} must not touch args {id:2}.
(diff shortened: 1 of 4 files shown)


prisis:
Superseded — RPC-error redaction landed via #70 (plan 064) and keyed optimistic fan-out via #59 (plan 065), both merged to alpha.

PR #57, opened 2026-06-30
security(do): redact unhandled RPC error messages on the lfse line

prisis:
## What

Brings the `errorToResponse` redaction to the `lfse-followups` integration line. The generic `RPC_FAILED` 500 path in `@lunora/do`'s `ShardDO.errorToResponse` echoed the raw `error.message` to clients — an unhandled throw can carry SQL fragments, file paths, or internal identifiers (Sensitive Information Disclosure).

Now it logs the raw error server-side (`console.error`) and returns a generic `{ code: "RPC_FAILED", message: "internal error" }`, mirroring `@lunora/runtime`'s `toErrorResponse`, which already redacts.

## Why a separate PR

The identical fix is already queued for `alpha` in #56 (`security/redact-rpc-errors-alpha`, open + mergeable). The `lfse-followups` line is a separate stack that still ships the leak until it next syncs from alpha, so this closes the gap on the active line now. Cherry-picked from the same commit, so the content is identical and will reconcile cleanly when the lines converge.

## Verification

- `@lunora/do` build + `shard-do.test.ts` (56 tests, incl. a new redaction-regression test asserting a sensitive message is not echoed) ✅
- types + eslint + prettier ✅

### DIFF
--- a/packages/do/__tests__/shard-do.test.ts
+++ b/packages/do/__tests__/shard-do.test.ts
@@ -218,7 +218,7 @@ describe("shardDO", () => {
         expect(response.status).toBe(400);
     });
 
-    it("returns 500 with mapped error when handleRpc throws", async () => {
+    it("returns 500 with RPC_FAILED and generic message when handleRpc throws", async () => {
         expect.assertions(2);
 
         shard.rpcResult = undefined;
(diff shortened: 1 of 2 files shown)


prisis:
Superseded — RPC-error redaction merged to alpha via #70 (plan 064).

PR #62, opened 2026-06-30
fix(do): emit settled frame on suppressed list result so the optimistic overlay drops

prisis:
**Plan 068** (bug, P1). Fixes the `@lunora/db` optimistic overlay hanging forever when a confirmed mutator's authoritative result does **not** change the subscribed list (a no-op update, or a write to a row filtered out of the list). The DO suppresses a byte-identical list frame, so `onRows` never fires and `awaitMutationId` never resolves.

**Server-change fix (the `settled` frame):** in `pushSubscriptionData`, the suppression branch now emits a rows-free `{type:"settled"}` frame (carrying the CDC cursor/epoch + per-client `lastMutationId`) instead of returning silently. The client's new `handleSettledMessage` advances the cursor/epoch and fires a new list `onCheckpoint` — **without iterating row callbacks**, so no re-render. `@lunora/db` wires `onCheckpoint` on the list path to resolve the gate from `confirmedMutationWatermark()`.

This is the live-write analog of the existing reconnect `resume` frame and the shape path's `onCheckpoint`. **Flicker is eliminated by construction** — there is no ack-based backstop; the gate advances only on a frame (data → `onRows` emits-then-resolves, or settled → `onCheckpoint` with nothing to render).

- Spans `@lunora/do` + `@lunora/client` + `@lunora/db` (9 files).
- Gates green: do 901 · client 261 · db 37; lint:types + eslint clean. Trigger-based wiring proven by a removal test.
- Tech-lead review: ✅ APPROVED. **Supersedes** the earlier client-only attempt (`advisor/068-db-list-overlay-hang`), whose ack-backstop flickered the changed-result path — do not merge that branch.
- Cost trade-off: a suppressed sub now sends a ~30–60 byte frame (only when its dependency tables changed but the result didn't); the expensive payload stays suppressed.


prisis:

PR #66, opened 2026-06-30
perf(do): share op-log read across shape pokes in one flush

prisis:
**Plan 072** (perf, P2). Shares the op-log read across shape pokes in one flush: extracts `collapseOpRange` and adds a per-flush `(table, sinceSeq)` cache so every shape resolving the same range reuses the collapsed op map. The per-shape membership probe stays per-shape.

- Tech-lead review: ✅ APPROVED — full do suite 908 green incl. the 070 matrix.
- **Stacked on `advisor/070`** (this PR targets that branch so the diff is 072-only).


prisis:

PR #67, opened 2026-06-30
perf(do): dedup identity-independent reactive query runs across sockets

prisis:
**Plan 073** (perf, P2). Dedups identity-independent reactive query runs across sockets in one flush — the handler executes once per flush for admin/reserved reads instead of once per socket. Per-socket push still runs.

**Security boundary:** the identity-independence signal is `isAdmin` (`ADMIN_FUNCTION_PREFIX`) **only** — `FLAGS_FUNCTION_PREFIX` is explicitly excluded and all user queries are treated identity-dependent. The per-identity path is byte-for-byte unchanged.

- Tech-lead review: ✅ APPROVED — verified the predicate can never be true for an identity-reading query; includes a leak-detecting test (Case 11) and a forced-predicate sanity-check.
- **Stacked on `advisor/072`** (this PR targets that branch).


prisis:

PR #68, opened 2026-06-30
refactor(do): extract shared socket-pool helper

prisis:
**Plan 074** (tech-debt, P3). Extracts a shared `runSocketPool` helper, unifying the two hand-rolled bounded worker-pool loops in `refreshSubscriptions` and `pokeShapeSubscribers`.

- Tech-lead review: ✅ APPROVED — behavior-preserving. The poke-path `await` yields are unobservable (the pool drains within one macrotask, preserving order and atomicity vs. external events); 910 tests = baseline.
- **Stacked on `advisor/073`** (this PR targets that branch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)
https://claude.ai/code/session_01JGzpcgGkQSQpbCd1nJLpTx

PR #58, opened 2026-06-30
security(do): redact raw error.message in RPC error fall-through

prisis:
**Plan 064** (security, P1). Redacts the raw `error.message` in the DO RPC error fall-through so internal error text can't leak to clients; a generic message is returned instead.

- Tech-lead review: ✅ APPROVED — scope clean, done-criteria re-run green.
- Part of the Wave 4 sync-engine hardening batch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)
https://claude.ai/code/session_01JGzpcgGkQSQpbCd1nJLpTx

## Summary by CodeRabbit

* **New Features**
  * Added local-first sync support with partial replication views and optimistic custom mutators across the client, server, and app libraries.
  * Introduced new framework hooks/composables for mutator handling in React, Vue, Solid, and Svelte.
  * Added new DB collection and mutator subpaths for easier integration.

* **Bug Fixes**
  * Improved conflict handling for concurrent edits and offline replay.
  * Added safeguards for invalid shape/table references and cross-shard join cases.

* **Documentation**
  * Added guides and examples for local-first sync, shapes, and custom mutators.

PR #70, opened 2026-06-30
fix(do): wave-4 shard-do hardening (redact errors, settled frame, dedup, op-range cache, socket-pool)

prisis:
Re-execution of the Wave-4 shard-do plans **onto current `alpha`**, as one coupled stack (all five edit the same `packages/do/src/shard-do.ts` regions sequentially, so they can't be independent PRs). Each carries its thermo-nuclear review fixes folded in. Tech-lead reviewed; all 6 gates green locally (`@lunora/do` 910 + `@lunora/client` 262 + `@lunora/db` 35 tests; types clean).

**Commits (merge with a merge-commit to preserve all five for semantic-release):**

1. `fix(do): redact raw error messages from rpc/ws error frames` (plan 064) — `errorToResponse` RPC_FAILED fall-through + `handleStream` catch-all log server-side and return generic `"internal error"`; coded/structured errors keep their messages.
2. `fix(do,client,db): emit settled frame to drop stale list overlays` (plan 068) — DO emits a `settled` frame on the byte-identical suppression path, gated on `socketClientWatermark(ws) !== undefined` (clientId only). Client parses via shared `ackAndAdvanceCursor`; `@lunora/db` wires `onCheckpoint` on the list path. Dead `lastMutationId` removed from Data/Resume messages.
3. `perf(do): dedup identity-independent reactive query runs` (plan 073) — flush-local memo collapses N identical admin/reserved reactive runs to one. `isIdentityIndependent` is `ADMIN_FUNCTION_PREFIX`-only (security boundary); RLS/`ctx.auth`/per-user-flag reads bypass the cache. Negative + force-override sanity tests.
4. `perf(do): share op-log read across shape pokes in a flush` (plan 072) — flush-local `(table,sinceSeq,upTo)` cache shares one `__cdc_log` drain across same-range shape diffs; read-counter test.
5. `refactor(do): extract bounded socket-pool fan-out helper` (plan 074) — extracts the duplicated bounded loop into `runSocketPool`; behavior-preserving; dedicated test.

Supersedes #58, #62, #66, #67, #68 (stale pre-divergence executions).

## Summary by CodeRabbit

* **New Features**
  * Added checkpoint notifications for subscriptions, helping clients react when writes are confirmed.
  * Improved subscription handling for settled updates so the app can recognize confirmed changes even when no new data is returned.
  * Increased concurrent socket processing for faster handling of multiple tasks.

### DIFF
--- a/packages/client/__tests__/lunora-client.test.ts
+++ b/packages/client/__tests__/lunora-client.test.ts
@@ -456,6 +456,110 @@ describe("lunoraClient", () => {
             expect(last).toEqual({ id: sub.id, type: "unsubscribe" });
         });
 
+        it("forwards a settled frame's watermark to onCheckpoint without firing the data callback", () => {
+            expect.assertions(3);
+
+            const client = new LunoraClient({
+                clientId: "client-A",
+                fetch: vi.fn<typeof fetch>(),
+                url: "https://app.example",
+                WebSocket: createMockWebSocket(),
+            });
+
+            const received: unknown[] = [];
+            const checkpoints: { checkpoint?: number; mutationId?: number }[] = [];
+
+            client.subscribe(fnRef("messages:list"), {}, (d) => received.push(d), {
+                onCheckpoint: (watermark) => checkpoints.push(watermark),
+            });
+
+            const socket = latestSocket();
+
+            socket.open();
+
+            const sub = firstSub(socket);
+
+            socket.receive({ id: sub.id, type: "ack" });
+            // A write touched the read tables but the result was byte-identical, so
+            // the server sent a `settled` frame instead of a data frame.
+            socket.receive({ cursor: 12, epoch: "e1", id: sub.id, lastMutationId: 5, type: "settled" });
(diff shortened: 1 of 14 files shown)


prisis:
Addressed both CodeRabbit findings:
- **Major (subscription.ts)** — `onCheckpoint` is now a `checkpointCallbacks: Set` on the shared `SubscriptionState`; `subscribe()` registers/unregisters per subscriber and `handleSettledMessage` fans out to all, so a `@lunora/db` collection joining a query a plain `useQuery` already opened still gets the overlay-drop signal. Added a regression test.
- **Minor (socket-pool/pokeShapeSubscribers)** — wrapped `pokeOne` in a per-socket `try/catch` mirroring `refreshOne`, so one socket's `awaitWsDrain`/`sendPoke` rejection can't abort the fan-out to its siblings (memos left intact → next flush retries).

Branch rebased onto current alpha; fixes folded into the 068 and 074 commits.

PR #73, opened 2026-06-30
feat(do): plan 075 Phase 1 — fan-out observability

prisis:
## What

Implements **Plan 075 Phase 1** (observability only) — the "you can see it scale" half of the auto-elastic fan-out relay tier, with **no behavior change**. Plus carries the stranded presence-phasing doc correction (see below).

## Why

Plan 075's relay tier promotes a topic/shape once its per-flush fan-out cost (O(subscribers)) becomes the bottleneck. Phase 0 needs that threshold grounded in **real numbers**, and the DX goal is that Studio *shows* a topic getting hot. Phase 1 makes the cost measurable before any topology change exists. Phases 2–4 (the transport/RLS rewrite) stay gated on Phase 0 sign-off.

Builds on plans **072** (shared op-range) and **073** (identity-independent runs), now landed on alpha.

## Changes

**`@lunora/do`**
- `pokeShapeSubscribers` + `broadcastWhisper` record per-pass fan-out counters: passes, sockets iterated, sockets delivered, peak width, and — for the async poke path only — coarse wall-time. In-memory, reset on hibernation, sharing `metrics.sinceMs`.
- New `__lunora_admin__:getFanoutMetrics` read folds **live per-topic/shape subscriber counts** from socket attachments (`summarizeFanoutTopics`) and returns them with the running counters.
- `recordFanoutPass` is pure (returns updated counters); instrumentation never changes which sockets are poked or who receives a whisper.

**`@lunora/studio`**
- A new **Fan-out** page under Observability (modeled on the subscriptions panel): hot topics by subscriber count + per-path cost cards.
- Wire types hand-mirrored in `lib/admin.ts`, with key **drift-guards on both the `@lunora/do` and `@lunora/studio` sides**.

**Design notes**
- Counting is **always-on** — integer increments are negligible; no env flag needed. The admin read is already `LUNORA_ADMIN_TOKEN`-gated.
- Socket-count **width** is the exact, reliable cost signal (the plan's O(subscribers) model). `totalMs`/`maxMs` are **coarse** (a DO clock advances only on I/O) and captured for the async poke path only — omitted for the synchronous whisper broadcast rather than shown as a misleading `0`.

## Tests & verification

- Pure `summarizeFanoutTopics` / `recordFanoutPass` unit cases + an end-to-end `getFanoutMetrics` admin-read assertion folding real socket attachments.
- Wire-shape drift guards (compile-time + runtime) on both packages.
- Full `@lunora/do` suite green (**938 passed**); `@lunora/do` + `@lunora/studio` typecheck clean (the 6 remaining studio `tsc` errors are pre-existing auth-panel baseline, confirmed against pristine alpha); ESLint clean on all changed files (one pre-existing `tabLabel` react-hooks warning, not introduced here).
- Studio jsdom component tests aren't run here (sandbox limitation); the panel is verified via typecheck + lint and the server contract via the DO integration test.

## Docs

- Carries the **presence-phasing correction** (originally #55, stranded on the diverged `lfse-followups` base): `usePresence` is a `listPresent` reactive query, so it belongs in Phase 3 (RLS-uniform), not Phase 2 (whisper). All touch points reconciled.
- Marks plan 075 **Phase 1 SHIPPED** in the plan + README status row.

Supersedes #55 (whose base branch diverged from alpha).

https://claude.ai/code/session_019NXayJPpMpVhqzEd4dLEx4

### DIFF
--- a/packages/do/__bench__/fanout-whisper.bench.ts
+++ b/packages/do/__bench__/fanout-whisper.bench.ts
@@ -1,0 +1,112 @@
+import { beforeAll, bench, describe } from "vitest";
+
+import type { ShardDOState } from "../src/shard-do";
+import { ShardDO } from "../src/shard-do";
+
+/**
+ * Whisper fan-out cost vs subscriber count — the regression guard for plan 075's
+ * O(subscribers) per-flush fan-out (the cost the auto-elastic relay tier targets,
+ * and that `getFanoutMetrics` surfaces).
+ *
+ * Each iteration drives ONE real `broadcastWhisper` through the public
+ * `webSocketMessage` path: the owner scans every connected socket, reads its
+ * hibernation attachment, and sends the one shared frame to topic members. The
+ * bench measures that loop at a small and a large subscriber count so a
+ * regression (e.g. building the frame per-socket, or an O(N²) membership check)
+ * shows up as a super-linear jump between the two.
+ *
+ * Harness notes. `getWebSockets()` returns ONLY the N topic members, so the
+ * fan-out iterates exactly N — the senders live in a separate pool (they need no
+ * membership and `broadcastWhisper` reads the sender only via the `sender`
+ * argument), keeping the iterated count equal to N rather than N + senders.
+ * Senders rotate through a large pool so no single sender exceeds the per-socket
+ * whisper token bucket (`WHISPER_RATE_BURST`), which would otherwise make a
+ * measured call early-return instead of fanning out. `send` is a no-op counter,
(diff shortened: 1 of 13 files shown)


prisis:
@coderabbitai review

prisis:
@coderabbitai full review

prisis:
Consolidated into #78 — a single working branch (`feat/075-fanout-relay`) for plan 075, rebased on the latest alpha. All commits from this PR are in #78; continuing the work there.

PR #90, opened 2026-07-02
fix(security): remediate the run-1 audit — RLS, fan-out, payments, SSRF, WS-hijack + hardening

prisis:
## Summary

This branch bundles **two independent run-1 security-audit passes** over the framework. The sections immediately below (**Pass A** — HIGH/MEDIUM/LOW/Latent bug) are the first pass. A **second pass** (added in commit `close run-1 audit findings`) closes a further **3 HIGH · 3 MEDIUM · 2 LOW** findings in the RLS engine, the fan-out path, the payment/flags/mask surfaces, and the browser SSRF guard — see the **Second audit pass** section near the end.

Across both passes the two hardest boundaries — RLS and SQL construction — were already fundamentally sound; these fixes close the gaps around them and in the opt-in add-ons.

### Pass A

Remediates the findings from a deep security audit of all 40 packages (run-1 baseline). One HIGH, four MEDIUM, sixteen LOW, one latent codegen bug, plus a code-drift cleanup.

## HIGH

- **Cross-Site WebSocket Hijacking** — `/_lunora/ws` did no `Origin` check and rode the victim's cookie, bypassing the HTTP CSRF guard (which exempts GET). Added `enforceWebSocketOrigin` (runtime); cross-origin cookie-bearing upgrades are rejected; same-origin, `csrf.trustedOrigins`, and token/no-cookie clients pass.

## MEDIUM

- **Cross-tenant shard access / fan-out open by default** — now **default-denied (403)** without an `authorize*` callback. Opt back in with `allowUnauthenticatedShardAccess: true` (single-tenant apps relying solely on per-row RLS).
- **Session cookie `Secure` flag** — `create-auth` now defaults `useSecureCookies` **on** unless `baseURL` is explicitly `http://` (Workers serve HTTPS in prod; unset `baseURL` no longer ships insecure cookies).
- **MCP boundary** — read-only by default (`allowWrites` / `LUNORA_MCP_ALLOW_WRITES`); write tools are omitted from the tool list **and** refused at dispatch; every run tool is allowlisted against discovered functions and kind-checked.

## LOW (hardening)

CORS predicate+credentials warning · conservative default SSR HTML CSP · generic admin errors + admin-gate-before-method-check · opt-in relay control-channel HMAC (`LUNORA_RELAY_SECRET`) · conservative cohort-multicast mask gate · `contains` LIKE wildcard escaping · payment redirect scheme validation · wider client identity fingerprint · R2 SQL identifier allowlist · hyperdrive `ctx.sql` injection docs · CLI template symlink skip · bounded storage signed-URL key cache.

## Latent bug

`ctx.flags` codegen now routes through the `lunorash` umbrella (`lunorash/flags`) instead of a hardcoded `@lunora/flags`, so umbrella-only apps no longer emit an import for a package they never declared. Regression test added.

## ⚠️ Behavior changes (intentional)

- Apps that used client-chosen shard keys / fan-out **without** an `authorize*` callback must now set `allowUnauthenticatedShardAccess: true`.
- Cross-origin cookie-authenticated WebSocket connections are now blocked.
- `useSecureCookies` now defaults on for non-`http://` deployments.

## Testing

Rebased on latest `alpha`, rebuilt all 40 packages. Green: runtime 436, do 992, codegen 483, auth 134, client 291, cli 519, bindings 206, react 98, storage 81, mcp 32, sql-store 29. Full repo ESLint (43 projects) and Prettier pass. Adds reject-case/regression tests for H1, M1/M2, M4, L7, L8, and the flags-umbrella bug.

Deferred (documented, not defects): admin WS token in query string, browser DNS-rebinding, mail policy, template branch pinning, `quoteIdentifier` triplication, and the larger structural refactors (unify `LunoraError`, `ctx.*` capability descriptors, `codegen --check`, splitting `shard-do.ts`).

---

## Second audit pass (commit `close run-1 audit findings`)

An independent audit pass, hunted → adversarially validated → verified against source. 3 HIGH · 3 MEDIUM · 2 LOW, all with regression tests.

### HIGH

- **Fan-out relation confused-deputy** — the reserved `__lunora_relation__:*` fan-out authorized on `envelope.fanOut.table` but the shard read `envelope.args.table`, an independently client-controlled field. A client whose app scopes `authorizeFanOut` to a table could authorize a decoy and read a *different*, sensitive table — a raw, RLS-blind cross-tenant dump. `parseEnvelope` now binds the read table to the authorized table and rejects a mismatch (`runtime/create-worker.ts`).
- **Write-only RLS policy opened unfiltered reads under `.rls("required")`** — a protected table appearing in the bundle only via a write (insert) policy routed reads to the *unguarded* writer with no filter, returning every row and silently defeating the secure-by-default guard. Reads now route to the raw writer only when an active `on:"read"` policy exists; otherwise the guarded writer fails closed. Same fix on the `get` by-id path (`server/rls/middleware.ts`). Mirrors the fail-closed parity already in `shape-read-base.ts`.
- **Unpaid `incomplete` subscriptions granted paid entitlements** — Stripe/Polar mapped `incomplete` (first payment not completed) → `trialing` → `active` → entitled. With Stripe's recommended `payment_behavior: "default_incomplete"` this entitled anyone who abandoned checkout, for ~23h, repeatably. `incomplete` now maps to non-entitling `past_due`; `checkout.session.completed` (subscription mode) is gated on `payment_status` (`payment/providers/{stripe,polar}.ts`).

### MEDIUM

- **SSRF guard bypass via trailing-dot FQDN** — the Browser Rendering private-host denylist missed `http://redis.internal./` / `http://metadata.google.internal./` (the WHATWG URL parser keeps a trailing dot on named hosts, strips it on IP literals). Strip a trailing dot before classification (`browser/create-browser.ts`).
- **`mask()` value oracle** — masking redacted output values but let a caller filter/probe by a masked column (`where: { ssn: { eq: X } }`, range binary-search, or a `.filter` predicate over the unmasked row) to recover the hidden value. Now rejects a masked column referenced in `where`/`count`, and masks the row before `.filter` predicates (`server/mask/middleware.ts`). *(Residual: the narrower index-ordering oracle needs schema index metadata the mask layer doesn't yet receive — documented, not bodged.)*
- **Public feature-flag channel oracle/spoof** — the public `useFlag` WS channel evaluated any client-supplied flag key (probing internal flags) and forwarded arbitrary client targeting context (spoofing attribute-gated flags). Now serves only discovered `LUNORA_FLAG_KEYS` and never honors client targeting context on the public channel (`codegen/emit.ts`).

### LOW

- **RLS write-denial table-name disclosure** — a policy-denied `patch`/`delete`/`replace` leaked the owning table name in the `FORBIDDEN` message. Dropped from the message; the rejection contract is unchanged (`server/rls/middleware.ts`).
- **Scheduled jobs could double-execute** — the scheduler dispatched with no dedup key despite a comment promising idempotency; a lost response after a committed side effect re-ran the job. The scheduler record id is now forwarded as `x-lunora-mutation-id` so the DO idempotency table dedups a re-fire (`runtime/create-worker.ts`).

Regression tests added across `runtime`, `server` (rls + mask), `payment` (stripe + polar), and `codegen`.

*(Handled during review: the fuller "indistinguishable-from-absent" form of the write-gate fix was scoped back to just the table-name leak because it broke the intended, tested `FORBIDDEN` rejection contract.)*

---

## Summary by CodeRabbit

* **New Features**
  * Added optional HMAC verification for internal relay control plus stricter relay masking rules.
  * Introduced MCP read-only vs write tool surfaces with explicit write enablement.
  * Added browser URL safety allowlisting and optional DNS re-checking for private-target blocking.
* **Bug Fixes**
  * Hardened security defaults across cookies, CSP/CORS/CSRF-origins, sharded access, RLS routing, masked “value oracle” queries, and SQL LIKE/escaping.
  * Improved R2 SQL identifier/table validation and tightened runtime/RPC fan-out protections.
  * Corrected payment subscription state mappings for “incomplete”/unpaid webhook cases.
* **Documentation**
  * Updated security and configuration guidance for the above changes.
* **Tests**
  * Added regression coverage for masking, authorization/CSRF, and security hardening behaviors.


prisis:
@coderabbitai /review

PR #138, opened 2026-07-11
feat(observability): grouped error Issues (local, OSS) — Phase 1

prisis:
## Phase 1 — Fingerprint + local Issues + container-crash grouping

First slice of the observability plan: **grouped error triage in local dev and any self-hosted worker**, shipping in the framework (pure OSS, no cloud dependency). Draws the error-grouping model from [superloglabs/superlog](https://github.com/superloglabs/superlog) (Apache-2.0). Phases 2–4 (OTLP transport, cloud ingest/store/hosted views, alerts+AI) stack later on the cloud branch (PR #85).

### What lands

**`@lunora/fingerprint` (new package)** — zero-dependency error-grouping core vendored from superlog. `fingerprintError({ functionPath, message, code? }) → { hash, title, culprit }`. The canonical hash is over `functionPath :: bucket(message)` only (code is metadata, never hashed), so a **live-sink error event and a persisted request-log row for the same error collapse onto one stable hash** — the same key a future cloud Incident groups on. The message bucketer strips per-request noise (URLs, UUIDs, IPs, timestamps, hex, long ids, paths) so a route-scanner sweep folds into a single Issue. Portable SHA-256 backend (workerd/browser-safe), proven byte-for-byte against upstream's `node:crypto` vectors. superlog copyright + `NOTICE` retained per Apache-2.0 §4; ships under the repo's `FSL-1.1-Apache-2.0`.

**`@lunora/do`** — new `__lunora_admin__:getIssues` admin RPC. `readErrorIssues` is a **pure read-side aggregation over the bounded `__lunora_reqlog__` readout** (no new storage, no transport): it groups recent `error`-outcome rows by fingerprint into `{ hash, title, culprit, count, firstSeen, lastSeen, sampleMessage }`, most-recently-active first. Wired through `readAdminOp` with an `ADMIN_WILDCARD` table dep so it serves both the one-shot POST and live WS subscriptions under the existing admin auth. Container lifecycle events now carry the **instance id** and parsed **exit code** onto `LogEntry`, and an error-level container event also appends an `error` request-log row — so **a container crash and a Worker throw fold into the same Issues list** under one fingerprint.

**`@lunora/studio`** — new **Issues** observability panel (live over the admin WS), threaded through the compile-checked route table. **Containers** moved from the Functions nav group into **Observability**, and the container fold is now keyed per `(name, instance)` with Instance + Exit columns so crash-looping instances are distinguishable.

### Verification

- `@lunora/fingerprint`: 19 golden-vector tests (bot-sweep collapse, live-sink↔reqlog hash equality, container-crash grouping) ✓ · tsc ✓ · eslint ✓
- `@lunora/do`: full suite **1042 passed + 1 todo** (incl. new `readErrorIssues` + container-crash-fold tests) ✓ · tsc ✓ · eslint ✓
- `@lunora/studio`: tsc ✓ · eslint ✓ · fold-container-instances unit suite 12 passed ✓
- Prettier clean across all changed/new files.

## Summary by CodeRabbit

* **New Features**
  * Added an Issues view in Studio to live-monitor and group recurring errors with filtering.
  * Introduced stable error fingerprinting to consistently aggregate identical issues across environments.
  * Enhanced container lifecycle monitoring with per-instance details and stop `exitCode` (including correlation `instance`).
* **Bug Fixes**
  * Improved issue coverage by recording crash-related lifecycle events and flushing durable request-log updates promptly.
* **Documentation**
  * Added documentation for error fingerprinting and populated license/notice materials.


prisis:
Addressed the review + bot findings in a284d9dc (plus a `/thermos` audit that surfaced two more):

**Bots**
- **CodeQL #25 (polynomial ReDoS, `superlog.ts`)** — replaced the regex stack-frame parser with a linear index-based scanner, and clamped the message bucketer's input length before its regexes run. The genuinely-reachable path in this PR was actually `messageBucketFor` (run per error row in `readErrorIssues`), not just the stack parser CodeQL pointed at — both are now defanged.
- **CodeRabbit (container flush)** — a container crash now marks the reqlog table changed and flushes, so live `getIssues`/`getLogs` refresh immediately. Same fix applied to the RPC error path, which had the same gap.
- **CodeRabbit ("per container" copy)** — reworded to "per instance" in `en.ts`, `studio.tsx`, and `containers-panel.tsx`.
- **react-doctor L54 / L104 / L107** — removed the redundant `useMemo`; Issue timestamps now format via the shared `formatTimestamp` helper instead of `new Date().toLocaleString()` in render.

**From the audit**
- `fold-container-instances.ts` had **literal NUL bytes** in the fold key, making git treat it as a binary blob (unreviewable diff). Replaced the hand-rolled delimiter with a JSON-encoded `[name, instance]` key.
- A container `stop` with a **non-zero exit code** never became an Issue (it arrives as `level: "info"`), contradicting the handler's docstring — now treated as a crash. Added tests for the non-zero-exit and clean-exit (`exit 0`) cases.
- Container crashes were persisted but **not emitted to Logpush** like every other error; extracted a shared `persistRequestLog` seam so the two writers can't drift.
- `readErrorIssues` now projects only the 3 columns it needs instead of the full 14-column hydrate + per-row `JSON.parse`, since the subscription re-runs on every write-flush.
- Added the `@lunora/fingerprint` row to the package table in `AGENTS.md`.

`@lunora/do` (1044 pass), `@lunora/fingerprint` (19), and the studio container/issues suites are green; tsc + eslint + prettier clean across the changed files.

PR #158, opened 2026-07-21
feat(observability): durable metric history, trend charts & trace exemplars

prisis:
Builds on the now-merged #149 (ctx.trace + ctx.metrics). Adds the durable/trend tier for `ctx.metrics`, rebased onto latest alpha.

## What's in it
- **Durable metric history** — a reserved SQLite table of per-minute `ctx.metrics` rollups on the ShardDO (mirrors `function-metrics.ts`: bounded retention, distinct-series cap, read clamp), surviving hibernation. New `getMetricHistory` admin RPC.
- **Trace exemplars** — every measurement is stamped with the recording dispatch's trace id (shard-owned); it rides onto the live buffer series and each history bucket.
- **Studio** — the Instruments table gains a **Trend** sparkline (from the durable history) and a **Trace** exemplar link that deep-links the Traces panel (one-shot `sessionStorage` handoff; `filterTraces` now matches `traceId`).

## Notes
- Squashed to one commit with the thermo-review fixes folded in (text-safe series key → also fixes the NUL bytes that shipped in #149's `metric-buffer.ts`, raw SQL handle so rollups don't pollute Query Insights, canonical `stableStringify` join key).
- Verified against latest alpha: DO 1163 tests, Studio 785 tests, both typecheck clean, prettier clean.

https://claude.ai/code/session_019M6G6CAoLVrQMxDYg7BWq2

## Summary by CodeRabbit

* **New Features**
  * Added durable per-minute metric history to power trend charts across sessions.
  * Added “trend” and “trace” columns to the Instruments table with sparklines and exemplar trace links.
  * Selecting a trace from Metrics now opens the Traces panel with the trace filter applied.
* **Improvements**
  * Trace search now matches full trace IDs (case-insensitive substring).
  * Metric exemplars now support drill-down from trends to the corresponding trace, when available.
* **Improvements**
  * Trend rendering is more resilient when history data is missing or incomplete.


prisis:
Pushed fixes for all review comments:
- **React Doctor** (sessionStorage in render, traces-panel): the exemplar hand-off is now read in a `useEffect` on mount, not the `useState` initializer.
- **CodeRabbit — carry the shard (Major):** `openTrace` now stashes `{ traceId, shardKey }`; the Traces panel seeds both its search and its shard on mount, so an exemplar from a non-root shard opens the right ring. Covered by a new test.
- **CodeRabbit — flat histogram mock (Minor):** `historyPoints` is now kind-aware (`sum = value * count` for histograms) so `checkout.ms` trends non-flat.
- Also rebased onto latest alpha (resolved the conflict) and regenerated api-snapshots.

prisis:
Ran a two-pass thermo review on the rebased branch and applied all findings:

**Code quality**
- **Structural (blocker):** moved the exemplar sessionStorage hand-off out of the *pure* `trace-geometry.ts` into a new `lib/trace-handoff.ts` built on the existing guarded `lib/browser-storage` accessors (a sibling of `lib/shard-history.ts`). `trace-geometry` is pure again and metrics-panel no longer imports a geometry module.
- **DRY:** collapsed `pointValue`/`metricHeadline` into one `projectByKind` so a series' headline and its sparkline point provably agree.

**Correctness**
- **Medium (hot-path write cost):** `recordMetricHistory` now fast-paths the in-minute repeat — an existing `(series, minute)` bucket costs one point-lookup + upsert, and the distinct-series cap scan + retention trim run only when a *new* bucket appears (≈ once per series per minute). Documented that a tight metrics loop should still pre-aggregate.
- **Low:** `readPendingTraceFilter` now type-checks `shardKey` too, so a tampered non-string can't reach the panel and crash `.trim()`.

Reviewers independently confirmed the three earlier review items are resolved, wire contracts are in lockstep, and exemplar trace ids are server-stamped only (not caller-injectable). Local: do 1163, studio 787, typecheck + api:check clean.

PR #185, opened 2026-07-25
perf(replica): add benches and speed up diff apply

prisis:
Surveys every package for CPU-bound hot paths that warranted benchmarks, adds suites where they were missing, and fixes the slow code the new numbers exposed.

## Benchmarks added

Five packages that had none: `@lunora/replica`, `@lunora/sql-store`, `@lunora/db`, `@lunora/advisor`, `@lunora/errors`. Each is wired exactly like the existing suites — `vitest.bench.config.ts` off the shared `getBenchConfig`, a `test:bench` script, `__bench__` in the tsconfig include, and the eslint test-file relaxations extended to benches. `vis run test:bench` now resolves 20 targets, so the CodSpeed workflow picks them up with no change to CI.

Packages deliberately left without benches: thin IO wrappers (`bindings`, `browser`, `hyperdrive`, `ai`, `mcp`, `x402`, `notify`, `queue`, `container`), build-time-only tooling (`config`, `vite`), and `studio` (UI). Benching those would measure the network or the harness, not our code.

## Performance fixes

| Change | Result |
|---|---|
| `applyDiffs` copied the row map once per diff — N+1 copies for N diffs — while its docstring claimed the opposite | **50x** faster on a 64-diff catch-up replay |
| `deriveInsertId` ran 64-bit FNV-1a over `BigInt`, allocating a heap object per character; now four 16-bit limbs in plain numbers | **~5x** on the hash, bit-identical digests |
| `v.array` built its result with `Array.from({ length })`, which has no V8 fast path | its own bench went from **6.46x slower** than the naive baseline to 1.05x faster |

The `v.array` case is worth calling out: `parse-hotpath.bench.ts` had been reporting the "optimized" parser as ~6x slower than the baseline it replaced. The number was sitting in the repo, correct, and unread.

## A correctness bug, found on the way

`@lunora/replica`'s canonicalizer sorted object keys with `localeCompare`, which resolves against the runtime's default locale and ICU version. Two clients in different locales derived **different** ids for the **same** row and diverged — precisely the failure `deriveInsertId` exists to prevent (REPLICA-05). Keys now sort by UTF-16 code unit, as canonical-JSON schemes specify.

**This changes derived ids** for id-less inserts whose keys sorted differently under the two orderings — broader than "mixed-case keys": punctuation and separators (`a-b` / `a_b`), and non-ASCII (`"é"` now sorts after `"z"`). Those ids were already environment-dependent, nothing in the repo persists them, and all three exports are `@experimental` — so this is a release note, not a migration.

## What the review changed

`/thermos` ran over the branch. Its most useful finding was structural: the hotpath bench measured two optimizations with one number. Splitting them apart showed that **the streaming canonical-JSON encoder I had written was not an optimization at all** — on its own it is ~2x *slower* than the canonicalized-copy + native `JSON.stringify` form it replaced, because `JSON.stringify` is a native fast path JS string building cannot beat. The end-to-end 3x was the limb hash carrying a slower encoder.

So the encoder is reverted. The rejected version stays in the bench as a labelled baseline, so the decision is re-checkable rather than folklore. Reverting it also deleted the entire `JSON.stringify`-parity risk surface (no hand-rolled escaping, no re-implemented `undefined`/`toJSON` rules) and with it a latent collision bug the review caught independently: pushing raw parts into an array meant a non-string `table`/`id` arriving from untyped wire JSON contributed *nothing* to the digest, so two distinct diffs derived the same row key. Interpolation coerces; a test pins it.

A second review round caught the same class of problem one level down: the benches were measuring hand-copied duplicates of the functions they guard, and the copy had already drifted. Exporting the real `fnv1a64Hex` / `canonicalizeForHash` and repointing benches and tests at them moved the hash claim from ~8x to a measured ~5x. That round also found the identical `localeCompare` determinism bug 200 lines away in `event-log-do.ts`, on the idempotency-fingerprint path — where two DO instances on different ICU builds would fingerprint one batch differently and reject a legitimate retry as a conflict. Fixed.

Also fixed from the review: a cast that defeated `noUncheckedIndexedAccess` for an unmeasured gain; two replica bench files that had drifted into measuring different inputs under near-identical titles (now a shared `apply-diff.shared.ts`); an advisor bench where a renamed lint would silently become a no-op *reporting as a dramatic speedup*; a `db` bench where ~5.7x of the `toMap` measurement was fixture construction; and several O(1) accessor benches that would only have produced noise-driven CodSpeed alerts.

One "optimization" was declined: `advisor`'s `circular_fk` looked like it dominated the suite, but only because the first fixture was a pathological 100-table FK ring. Johnson's algorithm is already the right choice there, so the *fixture* was fixed to be realistic and the ring kept as an explicitly-labelled worst case.

## Verification

1,167 tests pass across the touched packages, types clean, zero new eslint errors. The monorepo-wide affected run is green: 1,210 tests across 49 projects.

(`cli`'s `init-smoke` tests failed on an earlier run and passed on re-run — they fetch templates over the network and the vite template alone eats 6.4s of a 10s timeout. Worth a longer budget at some point; not touched here.)

---

## Measured under CodSpeed instrumentation

The `v.array` fix lands harder than the local numbers suggested — these are instrumented, not wall-clock:

| Benchmark | Base | Head | |
|---|---|---|---|
| `server` — 4-field mutation dispatch (optional + union) | 97.5 µs | 72.8 µs | **+34%** |
| `values` — `v.array(v.string())`, 32 items | 71 µs | 53.3 µs | **+33%** |
| `values` — `v.object()`, 7 fields with optional + array | 86.6 µs | 75.2 µs | **+15%** |
| `values` — nested object (user + message + record) | 115.7 µs | 104.1 µs | **+11%** |

The dispatch win is the one that matters most: every mutation validates its args through `v.array`, so this is on the request path, not just in a microbenchmark.

Also: 79 new benchmarks registered, 155 untouched.

### The one reported regression is not in shipped code

CodSpeed flags `parse-hotpath.bench.ts :: v.record — 16 entries :: baseline (proto-key skip + double path spread)` at −10.65%.

That is the **baseline arm** of a bench pair — an inline re-implementation living in the bench file, deliberately unoptimized, never shipped. It is byte-identical to `alpha` (`git diff origin/alpha...HEAD` touches neither `recordBaseline` nor `v.record`), and the only change to `packages/values/src/v.ts` in this branch is the three-line `v.array` edit.

The most likely cause is V8 code-layout: this branch adds `arrayFromPreallocationRegression` to that module, which can shift inlining and IC decisions for neighbouring functions. CodSpeed independently flagged the run with "different runtime environments detected", which points the same way. Production `v.record` is untouched and shows no change.

Needs an acknowledgement on CodSpeed rather than a code change — contorting a deliberately-naive baseline to satisfy a noise signal would defeat the point of having it.

## Summary by CodeRabbit

* **Performance**
  * Improved diff application, insert ID generation, fingerprinting, hashing, error conversion, SQL value handling, and schema analysis efficiency.
  * Reduced overhead when parsing arrays and applying batches of changes.
* **Reliability**
  * Ensured consistent key ordering for generated identifiers and fingerprints across environments.
  * Preserved stable handling of derived row IDs and unknown-row updates.
* **Testing**
  * Added comprehensive performance benchmarks covering common and high-load scenarios across core packages.


prisis:
All review comments addressed as of `aa2af8ea`.

| Comment | Resolution |
|---|---|
| `advisor:50` — `defineSchema(tables as never)` loses type safety | Fixed. Now `Record<string, TableDefinition>`, no cast. My earlier comment claiming a heterogeneous builder map couldn't be annotated was wrong — `defineSchema` is generic over exactly that type. |
| `advisor:109` — labels say 5/25/100 for 10/30/105-table schemas | Fixed. Labels derive from `schema.tables.length`, so they can't drift again. |
| `apply-diff-hotpath:74` — "current" arm measures the superseded canonicalizer | Fixed. `canonicalizeForHash` is now exported and the arm calls it. This also surfaced that the whole file was benching hand-copied duplicates — the copy had already drifted, which moved the hash claim from ~8x to a measured ~5x. |
| `apply-diff-hotpath:184` — hard-coded `500` | Fixed. Uses `ROW_COUNT` from the shared fixture. |
| `value-codec:117` — encode inputs allocated in the timed body | Fixed. Hoisted to module scope. |
| `parse-hotpath:22` — baseline no longer reproduces `Array.from` | Fixed. Added an explicit regression-guard arm; without it nothing in the repo exercised `Array.from({ length })` and the pessimization could have returned unnoticed. |
| `parse-hotpath:109` — guard allocates a path per element | Fixed. It now mirrors the shipped parser exactly except for array construction, which sharpens the reading to **6.7x** (6.1M vs 912K hz) — reproducing the ~6.46x this bench originally reported. |

Verification: 1,165 tests across the touched packages, types clean, zero eslint errors.

PR #189, opened 2026-07-25
feat(otel): close otel-cf-workers gaps and add wide events

prisis:
Closes the gaps between our OpenTelemetry support and [`evanderkoogh/otel-cf-workers`](https://github.com/evanderkoogh/otel-cf-workers), and adds a wide-event API for instrumenting a request without drowning in log lines.

## Why

An audit against `otel-cf-workers` turned up 11 gaps. Alpha has since closed two of them independently and better than the original plan did — resource attributes (incl. runtime detection) and inbound `traceparent` adoption — **so this PR builds on that work rather than duplicating it**. What remains:

| Gap | Fix |
|---|---|
| One HTTP subrequest per span/log/metric | Batching exporter, one request per signal per invocation |
| Only the `fetch` path traced | `queue()` / `scheduled()` / DO `alarm()` / socket frames all get a root span |
| Nothing below the dispatch span | Opt-out auto-instrumentation for `ctx.db` and `ctx.fetch` |
| No outbound trace context | `ctx.fetch` injects `traceparent` naming the span it records |
| Everything `SPAN_KIND_INTERNAL` | `kind`, plus span events, links and `recordException` |
| No redaction seam | `postProcessor` |
| Head sampling only | Real `tailSampler` — the buffer is what makes it possible |
| No `@opentelemetry/api` interop | New `@lunora/server/otel` subpath |

## The subrequest problem

`otlpSink` buffered nothing — every span, log line and metric was its own `fetch`. A Worker is capped at 50 (free) / 1000 (paid) subrequests per invocation, so a well-instrumented handler could spend the budget its own business logic needs. Now events buffer and ship as one request per signal, flushed at each invocation boundary via `ObservabilitySink.flush`.

Buffering is also what makes **tail** sampling possible: by flush time a trace has settled, so `tailSampler` can keep it on "anything failed or ran slow", and a dropped trace takes its correlated log records with it instead of leaving orphans.

A flush window is grouped by **resource** before wrapping — with `detectResources` on, two requests in one window can carry different host attributes, and an envelope has exactly one resource.

## Wide events (`ctx.span`)

```ts
ctx.span.setAttributes({ "user.plan": user.plan, "cart.items": cart.length });
const payment = await charge(cart);
ctx.span.setAttributes({ "payment.provider": payment.provider });
if (payment.retried) ctx.span.addEvent("payment.retried", { attempts: payment.attempts });
```

`ctx.span` is a handle on the span the request **already** emits. Facts accumulate across the dispatch and export as one OTel Event record (`event.name: "lunora.dispatch"`), so cost is flat however much you attach, and "p99 checkout latency for pro-plan users with >10 items" is one filter rather than a join across log lines.

It's a log record rather than span attributes because the DO's dispatch span shares its span id with the runtime's SERVER span — exporting both would put two partial spans with identical ids on the wire.

`ctx.log.event(name, fields)` adds the discrete-event counterpart (`LogRecord.eventName` **and** the `event.name` attribute, so old and new collectors both understand it).

## `ctx.db` instrumentation is tiered

Defaulting to on required a mode that can't blow up a trace:

- **`"summary"`** (default) — flat-cost counters (`db.calls`, `db.duration_ms`, per-op breakdown) folded onto the wide event. No new spans, no new records, cost independent of call count.
- **`"spans"`** — full waterfall, capped per dispatch, reporting `db.spans_truncated` rather than silently truncating.
- **`"off"`** — returns the db untouched (identity, not a proxy).

## `@lunora/server/otel`

`createOtelTracer(ctx)` returns a real OTel `Tracer` whose spans **are** `ctx.trace` spans, so a library instrumented against `@opentelemetry/api` (e.g. the AI SDK's `experimental_telemetry.tracer`) stops emitting into the Worker's no-op global provider. The two lifetime models are bridged by suspending the `ctx.trace` body on a promise that `end()` resolves.

No ambient context: tracking a current span across `await`s needs `AsyncLocalStorage`, which DOs avoid — so spans parent to the dispatch and the tracer is passed explicitly. Flatter than a full SDK, never wrong, never lost. Shipped as a **subpath, not a package**: 400 lines don't warrant their own release stream, and an **optional** peer dependency is how this repo already handles third-party integrations (`@lunora/ai`, `@lunora/payment`, `@lunora/hyperdrive`). Nothing installs `@opentelemetry/api` unless it imports the subpath, and the separate packem entry keeps it out of the main `@lunora/server` bundle. It's a *peer* rather than a dependency because the API keeps a module-global registry — a second copy would silently split it.

## Compatibility

No breaking changes. Batching is on by default (`batch: false` restores per-event POSTs); `traceFetch` and `instrumentDatabase` are opt-out; `ctx.span` / `ctx.log.event` are additive. Alpha's `resourceAttributes` bag keeps working and the loose 4-arg `wrapResource*` call sites are unchanged.

## Verification

Repo-wide `lint:types`, `lint:eslint`, `lint:prettier` clean; `pnpm run test` green across all 50 projects. New tests cover batching, tail sampling (including that a dropped trace takes its logs), post-processor fail-open, span kind/events/links, the traced fetch (header names the recorded span; query string stripped from `url.full`), db instrumentation tiering and its cap, wide events end-to-end via the harness, and the bridge.

## Summary by CodeRabbit

* **New Features**
  * Added richer tracing APIs: request-level `span` access plus span events, links, and exception recording; `SpanOptions`/`SpanKind` support.
  * Added `ctx.log.event(name, fields?)` for structured, named log events.
  * Added automatic database telemetry instrumentation (`off`/`summary`/`spans`) and outbound `fetch` tracing with optional `traceparent` propagation.
  * Improved trigger/worker observability with guaranteed telemetry flushing and wider event export; also supports traced queue/cron handling.
  * Enhanced OTLP exporting with buffered tail sampling and configurable post-processing.
  * Added testing-harness support for recording and retrieving wide dispatch span data.
* **Documentation**
  * Expanded bundled third-party license details (including OpenTelemetry/Apache 2.0).


prisis:
Addressed all review findings. Summary of what changed and why:

**🔴 `combineSinks` didn't forward `flush`** — correct and the most important one. This was a regression introduced when I ported the work onto `alpha`: the original branch had the flush fan-out, and re-authoring `otlpSink` onto alpha's `otlp-export.ts` structure lost it. Any batching sink wrapped through `combineSinks` — the pairing this module's own docs recommend — never received the invocation-boundary flush, so its buffer died with the isolate. `fanOut` now takes an argument list, because `flush(context)` puts the context first while the four `on*(event, context)` hooks put it second; a fixed shape would have handed `flush` an undefined context. Added two regression tests so it can't silently come back.

**🟠 `dispatchSpans` keyed by `traceId` alone** — correct, and a genuine data-loss bug. `resolveTraceAnchor` takes `traceId` from the inbound `traceparent`, so two concurrent RPCs under one client trace shared an entry: wide events merged and whichever dispatch finished first deleted the other's. Now keyed by `${traceId}:${rootSpanId}`. Your framing of *why* `traceSampling` is fine keyed by `traceId` — a sampling verdict is per-trace, a wide event is per-dispatch — is exactly right and is now the doc comment on the key helper.

**🟠 `"summary"" mode made every db-touching dispatch look like a wide event** — correct, and it defeated the root-span gate I'd documented as deliberate. `publishTally` wrote through `ctx.span`, which materialises the collector, so with `"summary"` as the default nearly every request fired the gate and exported a `lunora.dispatch` record for handlers that instrumented nothing. Took your first suggestion: the tally now lands on a separate slot on the dispatch entry that enriches a root span being recorded anyway, but never causes one. An uninstrumented handler emits nothing extra again.

**🟡 Head sampling silenced low-frequency triggers** — correct. Trigger events no longer ride the ratio at all; volume is inherently low and a ratio tuned for request traffic hid most fires of an hourly cron, which is the blind spot the wrapper exists to close.

**🟡 OTel `Exception` object form lost its message** — correct, applied your normalisation.

**🟡 Detached JSDoc blocks** — correct, all four reattached (member reordering for `member-ordering` had orphaned them). Also corrected two doc comments that had gone stale against the tally change.

**CodeQL: dead `ok` write** — correct. Declaring it `boolean | undefined` traded the dead write for a TS "used before assigned" error, so I removed the variable instead: it was exactly `error === undefined` on every path, since a non-2xx response sets `error` just as a throw does.

**Nitpicks** — both applied: extracted `sinkContextFor` so the two invocation boundaries share one shape, and removed the duplicated `parentSpanId` assignment in `otlp-export.ts`.

**One more found while reviewing your harness comment**, which was also right: the harness `recordException` omitted `exception.stacktrace`, so wide-event tests passed against a narrower record than production ships. Fixed with an assertion covering it. The remaining divergence is documented inline — production prefers a `LunoraError`'s stable `code` via `toErrorType`, which isn't exported from `@lunora/do`; for a plain `Error` the two agree.

Verification: repo-wide `lint:types`, `lint:eslint`, `lint:prettier` clean; `pnpm run test` green. Two `studio` `logs-panel` tests flake under the loaded parallel run — `packages/studio` is byte-identical to `alpha` on this branch and they pass 29/29 in three consecutive isolated runs, so that's pre-existing timing flake in `waitFor` + virtualization assertions, not a regression here.

prisis:
Correcting my earlier claim: I had **not** addressed everything. The 7 inline threads were fixed and are resolved, but I missed two later signals. Both are now fixed in `9f8e8ade`.

### 🟠 Security: redaction failed open

The third review flagged that `postProcess` returned the event **unmodified** when a hook threw. I'd written that fail-open deliberately and defended the reasoning — "a redaction bug should show as un-redacted data rather than as silence." That's right for observability in general and wrong for this hook specifically: `postProcessor` exists to strip PII before telemetry leaves for a third-party collector, and `error.message` routinely carries user input. A dropped span is a monitoring gap you notice; a leaked payload is an incident you can't retract. It now **fails closed**, and the test that asserted the old behaviour is inverted rather than deleted.

### ❌ Two CodSpeed regressions, both self-inflicted

I hadn't looked at the performance report at all. Both regressions were mine:

**`broadcastWhisper to 128 members` −31%** — I'd wrapped `webSocketMessage` in `withTriggerTrace`, which mints a trace anchor per frame (two `crypto` draws plus hex encoding) on a path with unbounded frequency. Removed there. The trade was poor regardless: a frame's work is a subscription re-evaluation already attributable to the RPC that wrote the data. `alarm` keeps the wrapper — low frequency, genuinely un-attributable background work, which was the blind spot actually worth closing.

**`rankPage hydration` −17%** — two causes. The db tally was published per call, rebuilding a `LogFields` object and doing a map lookup on *every query*; the shard now reads a plain counter object once at the dispatch boundary, so a query pays only integer increments. This is closer to what the original review suggested ("publish the tally once at the dispatch boundary") than my first attempt was. Separately, both `makeDispatchSpan` and the tally touched `dispatchSpans` eagerly on every `buildCtx` **even with no sink configured** — so a deployment exporting no telemetry was paying for it. Both now bail when nothing is collecting.

Please re-run CodSpeed to confirm; local wall-clock isn't comparable to its simulated instruction counts, so I'm not claiming the numbers, only that the per-call and per-frame work is gone.

Verification: repo-wide `lint:types`, `lint:eslint`, `lint:prettier` clean; `do` 1223, `runtime` 708, `server` 430, `codegen` 864, `testing` 90 all passing after merging the latest `alpha`.

prisis:
Swept every review signal on this PR against the current code. All 7 inline threads and both outside-diff findings were already fixed and resolved; **one nitpick was still genuinely open**, and it's now addressed in `a746e5be`.

### 🔵 Make tail-sampler failures observable and bounded

You were right that this one was invisible. `applyTailSampler` caught a throwing sampler, set `verdict = true`, and said nothing — so a broken predicate and a permissive one are indistinguishable at runtime: every trace is kept either way, and the only symptom is a collector bill nobody can account for.

Now reported to `console.error`, bounded on **both** axes:

- **Once per flush window.** The catch is per trace, so a window holding 50 traces would have emitted 50 identical lines. Failures are accumulated and reported once, carrying the first error and the affected trace count.
- **Five times per sink instance.** A sampler that throws throws on *every* window, so an unbounded diagnostic becomes exactly the log volume tail sampling exists to hold down — on a hot Worker it would outweigh the telemetry it's complaining about. The fifth line says it's going quiet. The counter lives on the sink rather than the module, so a fresh isolate reports again and a persistently broken sampler stays visible over time instead of being silenced permanently by the first burst.

### On fail-open vs. the configurable fail-closed policy

Kept fail-open, and wrote the reasoning down next to the catch instead of leaving it as a one-line assertion — because the PR now has the two hooks failing in *opposite* directions and that deserved an explanation:

`postProcess` fails **closed** because it guards PII, and a leaked payload can't be retracted. `tailSampler` fails **open** because it can only mis-judge volume, and dropping the trace deletes the evidence of whatever the sampler choked on. There's a sharper argument against fail-closed here too: a sampler that throws on the pathological traces — a cyclic attribute, a huge error message — would erase precisely the traces most worth keeping.

I didn't add the configurable fail-closed option. A config flag would imply `tailSampler` can be a volume *guarantee*, and it can't be: it's user code in the request path, and the only enforcement that survives a bug in it lives at the collector. The option's doc now says that outright rather than leaving a reader to infer a bound the hook can't offer.

Two regression tests: one asserts a 3-trace window reports once (not three times), one drives 12 flush windows and asserts it stops at 5 with the silencing notice.

**Verification:** `@lunora/runtime` 741 tests passing, `lint:types` / `lint:eslint` / `lint:prettier` clean. Extracted `groupByTrace` out of `applyTailSampler` along the way — the added branch pushed it past the `sonarjs/cognitive-complexity` ceiling.

The last CodeRabbit run was cut short by a review rate limit, so `9f8e8ade..a746e5be` hasn't been looked at yet — @coderabbitai full review when you have budget.

prisis:
CI is green apart from CodSpeed, and the two lint gates that were red are fixed in `9061c0dd`:

- **`Lint (package.json sort)`** — the `@opentelemetry/api` optional peer left `engines` above `peerDependencies`/`peerDependenciesMeta` in `packages/server/package.json`. Exactly the trap AGENTS.md documents: that job is the only thing in CI that sees key order, so it went green locally and red here. Pure key reorder, no value changes.
- **`Lint (api surface)`** — the wide-event work is a real public API addition that was never snapshotted (`SpanKind`, `SpanLink`, `SpanOptions`, `LunoraWideEvent`, across server/do/runtime/testing and the `lunorash` re-exports). Refreshed via `api:update`. I checked the diff for breakage rather than assuming: of 194 lines, only four are non-additive, and all four are backward-compatible widenings — `ContextTracer`/`LunoraTracer` widen their third parameter from `LogFields` to `LogFields | SpanOptions`, and the protected `recordUserLog` gains two *optional* trailing parameters. The no-breaking-changes claim holds.

## The CodSpeed report

**The two regressions I introduced are gone.** `broadcastWhisper to 128 members` (−31%) and `rankPage hydration` (−17%) are both back in the untouched set, so removing `withTriggerTrace` from `webSocketMessage` and moving the db tally to the dispatch boundary did what they were meant to.

**The one remaining regression is not from this PR.** `deep mixed stack` (−18.54%) lives in `packages/fingerprint/__bench__/fingerprint-stack.bench.ts`, and the benchmarked code is untouched:

- This PR changes **zero** files under `packages/fingerprint/`. `git diff origin/alpha HEAD -- packages/fingerprint/` is **empty** — the package is byte-identical to `alpha`.
- `@lunora/fingerprint` is genuinely zero-dep, as documented. Its only imports are `./sha256`, `./superlog` and `./lunora`. It reaches nothing in `shared/`, so the four `shared/*` files this PR does touch (`otlp.ts`, `otlp-batch.ts`, `log-event.ts`, `span-event.ts`) cannot be in its path.

CodSpeed footnotes the cause itself: *"No successful run was found on `alpha` (aa7abe6) during the generation of this report, so 49b10ad was used instead as the comparison base. There might be some changes unrelated to this pull request in this report"* — alongside the *"Different runtime environments detected"* warning. And `packages/fingerprint` is also unchanged between `49b10ad` and real `alpha`, so the substitute base doesn't explain it as a genuine delta either. Identical code measured across different runners is the whole story.

So this needs an acknowledgement on CodSpeed rather than a code change — I've left it rather than acknowledging it, since that's your call. Happy to re-run the benchmarks first if you'd prefer a clean comparison against a real `alpha` baseline before dismissing it.

PR #202, opened 2026-07-26
feat(codegen): emit api.mutators.* typed references

prisis:
## Summary

Reported back on [cameronapak/dotflowy#310](https://github.com/cameronapak/dotflowy/pull/310) after the post-#187 re-port: codegen still didn't emit `api.mutators.*`, so typed `serverRef` fell back to the string escape hatch (adoption feedback [finding #17](../blob/alpha/plans/170-dotflowy-adoption-feedback.md)). It wasn't one gap but a chain of three, all in `@lunora/codegen` + `@lunora/db`:

1. **Codegen never emitted the references.** `discoverMutators` lifted only the export name, so there was nothing for `serverRef` to point at — dotflowy's generated `ApiTypes` is literally `{}`. It now lifts each mutator's `args` validator map and its `server` impl return type, and `emitApi` renders `api.mutators.<name>: FunctionReference<"mutation", Args, Return>`. That's a real function reference, not a parallel shape: a mutator already dispatches through `LUNORA_FUNCTIONS` with `kind: "mutation"`, and the `namespace:fn` ref matches the `LUNORA_MUTATOR_PATHS` entry the DO's `isCustomMutator` override reads.
2. **`@lunora/db`'s reference overload could never match.** `R extends MutatorReference<never>` rejects any reference carrying a concrete arg type — `{ text: string }` is not assignable to `never` — so a genuine generated reference failed the constraint, fell through to the `serverRef: string` overload, and errored. The overload shipped in #187 was never exercised against a typed reference (its test used a bare `{ __lunoraRef }` literal), which is why the hole survived review. Now `<TArgs>(… serverRef: MutatorReference<TArgs>)`, inferring straight off the phantom.
3. **`_generated/server.ts` exported no `defineMutator`.** This is the cause of the 30-line hand-written `MutatorCtx` type and the **33** `const mctx = ctx as unknown as MutatorCtx` casts in dotflowy's `lunora/mutators.ts`: `defineMutator`'s `ServerContext` defaulted to the untyped base `MutationCtx`, and codegen re-exported project-typed `query`/`mutation`/`action`/`definePolicy`/`v` but not `defineMutator` — the one authoring surface whose `ctx.db` writes need the schema most. It now re-exports a project-bound `defineMutator`, and discovery accepts the `_generated/server` specifier so the typed path still registers.

Two latent qualifier bugs surfaced with the first checker-rendered mutator types, both pre-existing for ordinary functions:

- an extensionless `import("./_generated/dataModel")` qualifier (written by a function file that follows the repo's no-`.js` convention) emitted a **TS2835** into `_generated/*`, which is consumed under NodeNext;
- an `import("@lunora/values").Id<…>` qualifier was a **TS2307** for an umbrella-only app, which declares no `@lunora/values`.

And one more found by re-reading #310 (`fix(codegen): widen env for the defineNotify store factory`): `defineApp`'s `Env` is bound to `object` so a wrangler-generated `interface Env` is accepted, but `defineNotify`'s `store` takes `NotifyEnv` (`Record<string, unknown>`) — which an interface does not satisfy. **Every app with a `lunora/notify.ts` emitted an `app.ts` that fails tsc**, as `examples/notify-demo` does the moment it is regenerated.

Packages touched: `@lunora/codegen`, `@lunora/db`, docs (`apps/docs`, `@lunora/server` + `@lunora/db` package docs).

### Before / after

```ts
// Before — nothing checks the path, and the arg type is restated per mutator (×23)
insertSibling: defineMutator<{ id: string; userId: string; parentId: string | null; /* … */ }>({
    serverRef: "mutators:insertSibling",
    apply: (_ctx, args) => { /* … */ },
});

// After — bound at compile time, args inferred from the server mutator's validators
insertSibling: defineMutator({
    serverRef: api.mutators.insertSibling,
    apply: (_ctx, args) => { /* … */ },
});
```

```ts
// Before — `ctx` is the untyped base MutationCtx (×33 casts + a hand-written MutatorCtx)
import { defineMutator } from "lunorash/server";
server: async (ctx, args) => { const mctx = ctx as unknown as MutatorCtx; /* … */ }

// After — `ctx` is this project's typed MutationCtx
import { defineMutator } from "./_generated/server";
server: async (ctx, args) => { await ctx.db.patch(args.id, { text: args.text }); }
```

## Linked issues

- Refs `plans/170-dotflowy-adoption-feedback.md` #17 (and its new #17b / #17c)
- Refs [cameronapak/dotflowy#310](https://github.com/cameronapak/dotflowy/pull/310)

## Test plan

- [x] `pnpm --filter "@lunora/codegen" run test` — 891 passed (87 files), including new coverage for the mutator refs, namespace-collision precedence, mutator-free byte-identical output, both qualifier rewrites, and the notify env widening
- [x] `pnpm --filter "@lunora/db" run test` — 75 passed, including an `expectTypeOf` test asserting the **inferred** `args` type off a typed `api.mutators.*` reference (not just the runtime dispatch path)
- [x] `pnpm run test:affected` / `lint:affected:types` / `lint:affected:eslint` — green
- [x] `pnpm run lint:package-json`, `node scripts/api-snapshot.js check`, `prettier --check` on the diff — green
- [x] Golden fixture regenerated (`capture-expected`); `apps/playground` regenerated + `tsc --noEmit` clean, with its `lunora/mutators.ts` switched to the typed `defineMutator` so the new path is dogfooded
- [x] All 8 `examples/*` regenerated and typechecked individually

Reviewers may want to re-run the codegen + db suites and `examples/notify-demo`'s `tsc --noEmit` (that one fails on `alpha` once regenerated).

## Checklist

- [x] Updated relevant docs in `apps/docs` (or `README.md` for package-local docs)
- [ ] If a new package was added: `project.json` has `type:package` and a `category:*` tag — n/a
- [x] If migration impact: noted upgrade steps in the PR description and changelog entry — see *Migration* below

## Migration

Additive; nothing breaks. `serverRef: "mutators:sendMessage"` still works as the documented escape hatch, and importing `defineMutator` from `@lunora/server` / `lunorash/server` is still discovered. Two opt-in upgrades, both surfaced by re-running `lunora codegen`:

- point `serverRef` at `api.mutators.<name>` and drop the hand-written `defineMutator<{…}>` arg type;
- import `defineMutator` from `./_generated/server` and delete the local mutator-ctx type and its casts.

## Notes for reviewers

**Start with commit 1** (`feat(codegen): emit api.mutators.* typed references`) — the other four are consequences.

- The `@lunora/db` overload keeps a deliberate `@typescript-eslint/unified-signatures` disable. Merging the two overloads into a `string | MutatorReference<TArgs>` union leaves `TArgs` uninferable on the string branch, which then silently falls back to its default and erases the args type at every reference call site — the exact failure this PR exists to fix.
- `renderApiBody` now sorts members by export name. Discovery output was already sorted, so this is a no-op except where a namespace mixes discovered functions with synthetic entries (agents, mutators) — the golden `api.ts` fixture is unchanged.
- The last two commits are deliberately separate and reviewable on their own: `chore(codegen): regenerate example generated files` sweeps in the pre-existing drift the examples had accumulated against the emitter (notify-demo's `defineApp` / `AppTableName` blocks), and `chore(api): refresh the notify API snapshot` fixes drift from #195 that fails `api:check` on any branch that rebuilds `@lunora/notify` — no source change in either.

**Deferred, filed as #22–24 in `plans/170`** (found in the same re-read, each needs its own change):

| # | Finding | Evidence in #310 |
|---|---|---|
| 22 | Reads are owner-scoped declaratively (`defineShape({ owner: true })` from #9); writes got no counterpart | 23 mutators each declare a redundant `userId: v.string()` arg and open with a hand-rolled `assertOwner` |
| 23 | `bindMutators`' `collections` map is type-erased (`Record<string, Collection<Row, string>>`, invariant in the row type) | `nodes: collection as never` ×4, and all 23 `apply` bodies bypass `ctx.collections` to close over module state — the seam is dead code in its only production port |
| 24 | A fire-and-forget mutator has no error hook (the outbox path has `onWriteRejected`) | an app-level `trackLunoraMutation` wrapper on 8 call sites, purely to consume `isPersisted` rejections |

Also worth relaying to the adopter: `applyPlanToCollections` / `applyPlanToDb` (#18's shared-planner answer) shipped, so their hand-rolled `applyPlanToCollection` + `commitPlan` twins can go.

## Summary by CodeRabbit

* **New Features**
  * Added project-generated, typed support for custom mutators, including `api.mutators.<name>` references.
  * Introduced owner-scoped writes for mutators with fail-closed enforcement and rejection of forged owner values.
* **Bug Fixes**
  * Mutator renames/moves now surface as compile-time type errors instead of runtime failures.
  * Improved generated type/import handling for more reliable NodeNext consumption.
* **Documentation**
  * Updated mutator examples to use generated, schema-aware APIs and type-safe references.
  * Expanded guidance on `initMutators` and handling rejected writes via `onWriteRejected`.


prisis:
## Follow-up: a P0 found while auditing for other generation bugs

Pushed as `fix(codegen): accept umbrella imports in every discoverer`. Unrelated to the mutator work, but the same class as the two qualifier bugs — a hard-coded specifier list that nothing exercised.

Every discoverer hard-coded its own set of accepted module specifiers, and **three omitted `lunorash/server`** — the subpath the umbrella exists to provide, and the one codegen itself emits into `_generated/*` when a project depends on it. Discovery skips an unrecognized specifier **silently**:

| declared as | `@lunora/server` | `lunorash/server` (before) |
|---|---|---|
| `query` / `mutation` / `action` | registered | **dropped** — absent from `api.ts` *and* `LUNORA_FUNCTIONS`, so every call 404s at runtime |
| `cronJobs()` | registered | **dropped** — `_generated/crons.ts` not emitted at all; the schedule never fires |
| `defineMigration` | registered | **dropped** — `LUNORA_MIGRATIONS` empty; `lunora migrate up` finds nothing |

Codegen exits `ok` in all three cases.

Reproduced end-to-end against `apps/playground` (a real `node_modules`, so the checker resolves the callee symbol and actually reaches the gate) by switching one function file, a `crons.ts` and a `migrations.ts` to the umbrella form, then diffing the generated output.

**Why it hid.** With no resolvable package the checker has no symbol, so every discoverer falls back to matching the identifier text and the specifier gate is never consulted — which is exactly the configuration the fixture tests run in. Nothing in the repo imported `query`/`mutation` from the umbrella either (examples and the playground both use `./_generated/server.js`), so the one form an umbrella-only adopter reaches was the untested one. `discover-crons`' own comment already warned that a missed specifier "has every cron silently dropped" — and still missed the umbrella.

**Fix.** One shared `module-specifiers.ts` (`isServerSurfaceModule` / `isServerPackageModule` / `isCronSourceModule`), now used by functions, crons, migrations, mutators, shapes, env and identity discovery. The new `discover-module-specifiers.test.ts` installs a resolvable stub package into the fixture's own `node_modules` so the gate is genuinely exercised — **it fails on all five factories if the umbrella entry is removed** (verified by removing it).

Add-on factories (`defineQueue`, `defineWorkflow`, `defineAgent`, `defineContainer`) correctly stay scoped — the umbrella ships no subpath for them. `httpRoute` and `defineSchema` match on identifier text with no specifier gate, so neither was affected.

902 codegen tests pass; `test:affected` / `lint:affected:types` / `lint:eslint` / `api:check` green. Written up as finding #25 in `plans/170`.

prisis:
## All remaining findings fixed (#22, #23, #24)

Pushed as `feat(server,db): owner-scoped mutators, typed collections, rejection hook`. Every finding from the dotflowy re-read is now closed; `plans/170` marks #17/#17b/#17c/#22/#23/#24/#25 all FIXED.

### #22 — owner-scoped writes (`@lunora/server`)

`defineMutator` gains `owner: "<column>"`. Before `server` runs it:

1. **requires a verified identity** — an anonymous caller is rejected outright, the same fail-closed rule an `owner`-scoped `defineShape` applies to reads (filtering on a nullish value would match a nullable owner column);
2. **rejects a forged owner** — a client-supplied value that disagrees with the session is a `403`, not a correctable input;
3. **sets the column to the verified value** — so the impl reads `args[owner]` without trusting the client, and declaring it `v.optional(...)` drops it from the wire.

That deletes the `assertOwner(ctx, args.userId)` opening all 23 of dotflowy's mutators, and makes it impossible to forget on mutator 24.

**Design note worth reviewing:** it names the column instead of taking a shape's `owner: true`. A shape is bound to one `table`, so its `.ownedBy(field)` resolves unambiguously; a mutator may write several tables and has no single owning table to read it from. Supporting `true` would mean either threading a schema-derived `ownerField` through the generic `handler(ctx, args)` seam every registered function shares, or teaching codegen to discover `.ownedBy()` (`TableIR` doesn't carry it today) — real plumbing for one saved string literal. `contextUserId` moved to a shared module so reads and writes resolve identity identically.

### #23 — typed collections map (`@lunora/db`)

`ClientMutatorContext` / `ClientMutatorDef` / `BindMutatorsContext` are now generic over the map, plus `initMutators<TCollections>()` to bind `defineMutator` + `bindMutators` once (mirroring `initLunora.dataModel<DataModel>().create()`):

```ts
const { bindMutators, defineMutator } = initMutators<{ nodes: typeof wholeOutlineCollection }>();
```

`Collection` is invariant in its row type, so the old shared `Record<string, Collection<Row, string>>` did double damage: `as never` on every entry at the bind site, *and* an `apply` context too wide to use — which is why all 23 optimistic bodies ignored `context.collections` and closed over module-scope variables instead. Both ends are typed now; the standalone functions still work against the widest map.

### #24 — `onWriteRejected` (`@lunora/db`)

`{ args, code, error, mutator, serverRef }`, matching the outbox hook's shape. Setting it also **consumes** the `isPersisted` rejection, so a fire-and-forget call no longer needs an app-level wrapper (dotflowy's `trackLunoraMutation`, 8 call sites); a caller that awaits still observes it. Opt-in deliberately — an unconditional `.catch` would swallow dropped writes for callers who never asked. A throwing listener is swallowed so reporting a failure can't manufacture a second one.

### Verification

Everything is additive: a mutator without `owner` is untouched, and the standalone `defineMutator`/`bindMutators` keep their previous types.

- 451 `@lunora/server` tests, 80 `@lunora/db` tests (6 new owner cases, 3 new rejection cases, 2 new `initMutators` cases incl. `expectTypeOf` on both the collection row type and the inferred args)
- The `@tanstack/db` test mock gained a faithful `isPersisted` deferred that settles with `mutationFn` — the existing 16 cases pass through it unchanged
- `test:affected` / `lint:affected:types` / `lint:affected:eslint` / `api:check` / prettier green; all 8 examples typecheck
- API snapshots show exactly the intended surface (`owner?: string`, the generic params, `BoundMutatorApi`, `CollectionMap`, `MutatorRejectedEvent`)

Docs updated: a new "Owner-scoped writes" section in the local-first guide, plus `initMutators` / `onWriteRejected` sections in the `@lunora/db` package docs.

PR #213, opened 2026-07-27
feat(mcp): docs MCP server, `lunora mcp install` and `lunora mcp serve`

prisis:
Prompted by a look at [appwrite/mcp](https://github.com/appwrite/mcp), which does three things ours didn't: it serves the framework's **documentation** as a tool, it's **hosted** at a URL clients just point at, and its README is a **one-line install** for every editor. `@lunora/mcp` only exposed a deployed app behind an admin token — nothing helped the much more common case of an agent trying to write Lunora code in the first place.

Deliberately *not* copied: Appwrite's control-plane tools (create databases/buckets). Lunora's control plane is Cloudflare, which already has official MCP servers; re-wrapping wrangler is duplicated surface. Their OAuth flow has no analogue either — Lunora deployments are user-owned Workers with no central identity — so token-in-env stays right for the deployment server, and the *docs* server is the one that fits the hosted model, because it needs no auth at all.

## 1. `@lunora/mcp/docs`

`lunora_search_docs` / `lunora_get_doc` / `lunora_list_docs` over a `DocsIndex` interface. Free of credentials and of Node built-ins, so it runs on Workers and edge functions. Two backends satisfy the interface — a site's own in-process index, or `createRemoteDocsIndex` reading a published site's `/api/search`, `/llms.mdx/*` and `/llms.txt` — and both map results through one shared helper so a model sees identical hits either way.

The docs site mounts it at **`/mcp`**, backed by the existing fumadocs/Orama index (refactored into a shared `search-server.ts` so `/api/search` and `/mcp` don't each build one). Unauthenticated, CORS-open, and a plain browser GET returns a help page rather than a protocol error.

## 2. `lunora mcp install [client…]`

Knows each client's config file, top-level key and entry shape: `claude-code`, `cursor`, `vscode` (`servers`, not `mcpServers`), `gemini`, `claude-desktop` (per-OS path), `windsurf`, and `codex` (TOML → printed snippet). Edits go through `jsonc-parser`, so comments survive; existing entries are kept unless `--force`; a config that doesn't parse is reported, never overwritten. With no client named it installs into every client already configured in the project.

## 3. `lunora mcp serve`

The stdio server those entries spawn — docs tools + `lunora_dev_status` / `lunora_dev_logs` + the deployment tools. Config is `{\"command\": \"pnpm\", \"args\": [\"exec\", \"lunora\", \"mcp\", \"serve\"]}`: no env, no URL. The dev URL comes from `.lunora/dev.json` and the token from `.dev.vars`.

Two design points, both from things that broke in testing:

- **Deployment resolution is lazy, per tool call.** Editors spawn MCP servers when a project opens — usually before `lunora dev` runs — and keep them alive across restarts, so a URL captured at startup would be absent for the whole first session. The tools stay advertised either way (clients cache the tool list) and return an actionable error when nothing is running.
- **The first stdio run answered nothing.** The command framework exits with whatever the handler returns, and `connect` resolves as soon as the transport is *connected* — so the process died before its first request. `serve` now blocks until the server closes or stdin ends.

Supporting changes in `@lunora/mcp`: `createToolServer` to compose independent tool surfaces, `deploymentTools` to expose the existing surface as composable tools, and `serve-stateless` split out of `http` so the docs entry reaches the transport without importing the `package.json` read in `server`.

## Also: `lunora --version` printed `0.0.0`

Every published build. `readCliVersion` resolved `../package.json` relative to its own module — right from `src/`, wrong from `dist/`, because packem hoists that module into a hashed `dist/packem_shared/` chunk. It now walks up for the manifest whose `name` is `@lunora/cli`, the approach `@lunora/mcp` already uses; one code path covers vitest and any future re-chunk. (`@visulima/vis` pins its emitted depth, `../../package.json`, plus a `VIS_VERSION` override from its unbundled launcher; walking costs a stat or two and needs no launcher.)

No unit test can catch that class of bug — from `src/` the manifest is always where the resolver looks — so `dist:check` now runs the built binary and compares. Verified it fails on the old resolver and passes on the new one.

## Verification

`@lunora/mcp` 123 tests, `@lunora/cli` 719 tests; eslint (`--max-warnings=0`), prettier, `tsc --noEmit` and `vis sort-package-json --check` clean across `@lunora/mcp`, `@lunora/cli` and the docs app; docs app builds.

Beyond unit tests, I drove the real endpoints: the docs site's `/mcp` answered `initialize`, `tools/list`, a search for `shardBy`, `lunora_get_doc` on an origin+anchor URL, and `lunora_list_docs` returning all 138 pages; and the built CLI binary served a full stdio session and exited cleanly.

**Not verified:** the deployment tools against a live `lunora dev` — that needs wrangler and an example app. The resolution and client-caching logic is unit-tested and dispatch goes through the same `callTool` path the existing suite covers.

## Notes for review

- `apps/docs/src/routeTree.gen.ts` has churn beyond the new route — the generator reordered its imports.
- Fixing the version makes the update notifier live on the next publish. That's its designed behaviour (gated on CI, TTY, `LUNORA_NO_UPDATE_NOTIFIER`, 24h cache), but it is user-visible.
- Left alone: `lunora docs` opens `https://lunora.anolilab.dev/docs` while everything else, including the docs site's own `SITE_URL` and the MCP defaults here, uses `lunora.sh`. One line in `packages/cli/src/commands/docs/handler.ts` — say the word.

## Summary by CodeRabbit

* **New Features**
  * Added MCP support for docs, local dev-server tooling, and deployments, including a new `lunora mcp uninstall`.
  * Introduced `lunora mcp install`, `lunora mcp serve`, and `lunora mcp uninstall` for editor integration and stdio hosting.
* **New/Updated Web Docs**
  * Added an `/mcp` site route for the docs MCP surface over Streamable HTTP, with docs availability notice.
* **Documentation**
  * Expanded MCP guidance for installation, usage, and safer public hosting details.
* **Improvements**
  * Improved docs search and added request rate limiting to the docs MCP endpoint.


prisis:
Pushed fixes for all 7 review comments, plus findings from two local audit passes (bug/security and code-quality). Details in commit `46ae6df5`; merged `alpha` and regenerated the lockfile, so this is mergeable again.

### Review comments

| Comment | Fix |
| --- | --- |
| `serve-stateless.ts` — leak on failure | Teardown moved into a `finally`; a rejection from `connect`/`handleRequest` no longer leaks a server + transport per failed request. |
| `remote-index.ts` — unbounded request | `AbortSignal.timeout` (10s default, `timeoutMs` option). |
| `tools.ts` — bare slug | `"sharding"` now resolves to `/docs/sharding`, matching the documented equivalence it previously failed. |
| `local.ts` — ambiguous cache key | Serialized tuple instead of `|`-joined. |
| `install.ts` — `--print` vs skip | `--print` now predicts the skip the real write would take. |
| `routes/mcp.ts` — `Accept` case | Lowercased before the SSE check. |
| `ai-tooling.mdx` — "afterwards" | Fixed. |

### The one worth a second look

The unauthenticated `/mcp` endpoint accepted JSON-RPC batches, and the stateless transport buffers a whole batch's replies into one body. Measured against the built handler: **202 KB in → 87 MB out** across 2000 `lunora_list_docs` calls, with no `initialize`, no session, and no per-caller cost.

`createDocsMcpFetchHandler` now caps bodies at 128 KiB and refuses batches outright — batching buys a docs client nothing. Re-measured after: the 2000-item batch is a 413 with zero tools dispatched, a 50-item batch (under the size cap) is a 400 from the array guard, and a single call is unaffected. Query arguments and `lunora_list_docs` output are bounded too.

### Also fixed

- **`install` wrote a shape three clients can't use, and reported success.** Gemini CLI selects transport by property name (`httpUrl`; a bare `url` means SSE), Windsurf wants `serverUrl`, and Claude Desktop validates stdio entries only — it now gets the documented `mcp-remote` bridge.
- **`mcp serve --url` shipped the project's admin token to any origin.** Withheld across a non-loopback `--url` unless `--token` is passed.
- **`jsonc-parser`'s `modify` throws on a non-object server key**, and the throw escaped the client loop — one hand-edited `"mcpServers": null` aborted the install and silently skipped every remaining client. Gated and caught; config writes are also atomic now, since these paths include files we don't own.
- **Codex could never be auto-detected** (detection was gated on `format === "json"`).
- **The CLI was pulling in viem (53 MB), the Solana kit and the x402 packages**, via `@lunora/mcp` → `@lunora/x402`, for a paid-tools module `lunora mcp serve` never touches. `@lunora/x402` is now an optional peer, loaded lazily.

### Deliberately not done

- Unifying `createLunoraMcpServer` onto `createToolServer`. It would replace the fail-closed refusal messages ("this server is read-only, enable with…") with a bare "unknown tool", losing the signal that tells an operator *why*.
- Extracting the version resolver into `shared/`. It's duplicated with `@lunora/mcp`, but `shared/` consumers must drop `outDir`/`rootDir`, which isn't worth it for ~30 lines.

750 CLI tests, 133 mcp tests; lint, types and the dist guards clean.

prisis:
### Correction to my earlier comment

I wrote that `add-mcp` "independently arrived at the same per-client transport shapes." **That was wrong**, and I want it on the record before anyone relies on it. In `add-mcp`, `serverUrl` belongs to Antigravity, not Windsurf, and `httpUrl` appears only in a *reader* that scans candidate keys — never in a writer. It writes `{type, url}` for Gemini.

So I checked the primary sources instead:

- **Gemini CLI** documents `httpUrl` for Streamable HTTP, and states a bare `url` is an **SSE** endpoint.
- **Windsurf** documents `serverUrl` for remote HTTP servers.

CodeRabbit's finding stands on those docs — not on a second tool agreeing with it. Our shapes are correct, and `add-mcp` appears to have this wrong for Gemini: a server it installs there would be read as SSE and never connect.

### On reusing packages

Adopted two that were already in the tree, both replacing hand-rolled parsers that the review had found bugs in:

- **`smol-toml`** for the Codex snippet — wrangler already depends on it. Fixes the dot-in-server-name nesting bug and the unescaped control characters.
- **`es-module-lexer`** for the dist guard's chunk walk — already a catalog dependency. A real lexer beats regexes against minified bundler output.

**`add-mcp` itself: absorbed, not depended on.** Its client matrix is the genuinely valuable part, so `60fda1db` adds OpenCode, Cline and Zed from it (10 clients now). Every shape is labelled with its provenance — verified-against-docs vs adapted-from-add-mcp — precisely because the two disagree and only one of them is right. Taking the dependency would have added a second CLI framework and prompt library, and handed it the write path to `~/Library/Application Support/Claude/…` that this branch just made atomic.

PR #216, opened 2026-07-27
refactor(shard-engine): move the reactive engine core out of @lunora/do

prisis:
Plan 114 §5.2's remainder — the part #190 deferred as too large to review alongside it.

**Stacked on #190.** Base is `feat/platform-abstraction-layer`, because `@lunora/shard-engine` only exists there. Review after #190 lands; it will need one rebase onto the squashed commit.

## What moves

Twenty modules (~10,300 lines) from `@lunora/do` into `@lunora/shard-engine`: the `ctx-db` family, the CDC op-log, the relay tier, introspection, the system reader, triggers.

The closure was verified movable **before** anything moved:

| Check | Result |
|---|---|
| `@cloudflare/workers-types` imports | none in any of the 20 |
| Durable Object references | none |
| Edges back to modules that stay | none — self-contained |
| Cycle into `shard-do` | none |

`@lunora/do` keeps re-exporting everything. **Zero exported names lost** from `do` or `lunorash`, and `_generated/` goldens are byte-identical — the plan's hard constraint.

## The move exposed twenty duplicated types

This is the interesting part. The first extraction copied `ctx-db`'s type surface into `schema-types.ts` and left the originals in place. Two packages, two copies, structurally compatible at the boundary — so nothing complained. In **one** package TypeScript sees them as distinct identities (`DatabaseWriterLike$1`), and the duplication has to be resolved rather than tolerated.

Seventeen were byte-identical. The three that weren't each say something:

- **`SearchIndexDefinitionLike`** — the local copy carried `language`, `staged`, `strategy`; the canonical one did not. Alpha's #203 added those to `ctx-db`'s copy and `schema-types` silently never got them. **The canonical contract was stale against shipping code.** The richer shape is now canonical.
- **`DatabaseWriterLike`** — differed in exactly one member: `system?: unknown` vs `system?: SystemDatabaseReader`. The `unknown` was a cycle workaround — `system-reader` lived in `@lunora/do` while `schema-types` was already here. Both are in one package now, so the real type applies and callers stop asserting past it.
- **`ApplyOnDeleteOptions`** and **`GuardableSchema`** are genuinely different types sharing a name across modules. Left alone.

Same defect class as `ShardSqlExec` and the binding projections in #190: a second copy that nothing forced to agree drifts, and only wiring it up says so.

## Two things stayed behind deliberately

- **`relay-hub.test.ts`** needs `ShardDO` — it is an integration test and belongs with the DO, not the engine.
- **`_helpers` fixtures are duplicated**, not shared. The alternatives are exporting a test fixture from a published package's public API, or a cross-package relative import `rootDir` rejects. 123 lines of schema literal and a `node:sqlite` factory; each copy can drift toward what its suite needs.

## One bug worth naming

The moved tests now import relatively rather than through the package barrel. That is not cosmetic: importing the built barrel from *inside* the same package gave `ConflictError` two module identities, and four `instanceof` assertions failed against the class they were actually thrown from — `expected ConflictError … to be an instance of ConflictError`.

## Verification

| | |
|---|---|
| `@lunora/shard-engine` | 629 |
| `@lunora/do` | 687 + **46 workerd** |
| `@lunora/sql-store` / `@lunora/d1` / `@lunora/runtime` | 80 / 237 / 760 |
| totals | 1316 vs 1315 before — the relay-hub test re-counted after moving back |
| repo | uncached `tsc` on both packages, repo types, prettier → eslint → prettier, 65 manifests, API snapshots re-blessed |

## What this unblocks

The four engine-level TCK legs that plan 114 §3 specifies and §5.4 could not assert while the code implementing them was Cloudflare-bound: OCC snapshot-CAS → 409, poke `pokeStart`/`pokePart`/`pokeEnd` ordering with per-socket cursor resume, RLS identity under a live subscription, and scheduler at-least-once + dead-letter.


prisis:
**CI is not running the real matrix on this PR, and its 4 green checks should not be read as a pass.**

`lint.yml` and `test.yml` trigger on `pull_request.branches: [main, alpha, next, beta]`. This PR targets `feat/platform-abstraction-layer`, so only the always-on checks (CLA, semantic title, triage, CodeRabbit) fire — no lint, no types, no tests, no workerd.

Everything was therefore verified locally before pushing:

| | |
|---|---|
| `@lunora/shard-engine` | 629 |
| `@lunora/do` | 687 + 46 workerd |
| `@lunora/sql-store` / `@lunora/d1` / `@lunora/runtime` | 80 / 237 / 760 |
| types | uncached `tsc --noEmit` on **both** packages, plus repo-wide `lint:types` |
| format | prettier → eslint → prettier (eslint's autofix re-sorts imports past prettier's widths, so the third pass is load-bearing) |
| API | re-blessed; **zero** exported names lost from `do` or `lunorash` |
| goldens | `_generated/` byte-identical |

The matrix will run for real once #190 lands and this is retargeted at `alpha`. Worth re-running it then rather than trusting this page.

Issue #230, opened 2026-07-29
Decompose the ten oversized studio panels (react-doctor/no-giant-component)


prisis:
Ten `@lunora/studio` panels carry a `react-doctor/no-giant-component` suppression. Each was deferred deliberately during the react-doctor sweep in #229 — decomposing a panel is a real refactor with its own review, not something to do blind inside an unrelated change. This issue is the tracking record those suppressions point at, so the deferral is visible rather than invisible.

## The components

| Component | Lines |
| --- | --- |
| `src/features/sql/sql-editor-panel.tsx` | 839 |
| `src/features/data/data-browser.tsx` | 810 |
| `src/features/logs/logs-panel.tsx` | 800 |
| `src/features/schema/schema-viewer.tsx` | 669 |
| `src/features/reports/health-panel.tsx` | 640 |
| `src/features/auth/organization-detail.tsx` | 550 |
| `src/features/reports/metrics-panel.tsx` | 492 |
| `src/features/data/global-data-browser.tsx` | 452 |
| `src/features/logs/mail-panel.tsx` | 423 |
| `src/features/schema/schema-editor-overlay.tsx` | 380 |

Each site carries the line count and a pointer to this issue in its suppression comment.

## Why this is not just a lint number

These are the panels where a reader has to hold the most in their head at once, and they are also where the sweep found the most real defects — the SQL editor panel is where per-tab state had leaked into panel-wide state, and the data browser is where the render-phase ref writes and the identity-critical memos cluster. Size is the reason those were hard to see.

## Suggested approach

Decompose one panel per PR, not ten at once. The sweep already produced a usable seam for this: several `useMemo` bodies were hoisted to module-scope pure functions, and `use-data-browser.tsx`'s four (`toPageArgs`, `toCountArgs`, `resolveStagedChanges`, `editableColumn`) are pure, hook-free, and belong in a sibling `data-browser-args.ts`. Moving them would drop that file under the threshold, make them unit-testable without a renderer, and cost nothing behavioural — a good first PR.

For the panels themselves the split that tends to hold is: the data/orchestration hook (already extracted in the data-browser and file-browser cases), the toolbar, and the result surface. `packages/studio/src/features/data/` is the precedent to copy — `use-data-browser.tsx` + `data-browser-grid.tsx` + `grid-features.tsx` + `data-filters.tsx` around a thin `data-browser.tsx`.

## Definition of done

[...]

PR #231, opened 2026-07-29
docs(studio): cite the tracking issue in the giant-component suppressions

prisis:
Follow-up to the react-doctor sweep in #229, closing the one finding from the thermo review that could not be fixed in code.

The ten `react-doctor/no-giant-component` suppressions justified themselves with a claim that was not true. First they said the debt was **"tracked separately"** — nothing tracked it. The replacement said it was recorded under "Deferred" in `plans/README.md` — an entry that was never actually written. Same failure twice: the load-bearing half of the justification was fiction, which turns real debt into invisible debt.

#230 now tracks it properly: all ten components with their line counts, a suggested order, and a definition of done ("the suppression can be deleted and the scan still reports zero"). Each suppression cites the issue by number and carries its own exact line count. The Wave 15 notes in `plans/README.md` record the deferral so the plan history and the code agree.

The suggested first step in #230 is the cheap one the sweep already set up: `use-data-browser.tsx`'s four hoisted helpers (`toPageArgs`, `toCountArgs`, `resolveStagedChanges`, `editableColumn`) are pure and hook-free, so moving them to a sibling module drops that file under the threshold at zero behavioural cost.

Comments only — no code paths change. ESLint `--max-warnings=0`, `tsc --noEmit`, 930 studio tests, Prettier, and `react-doctor` (0 findings) all clean.

## Summary by CodeRabbit

* **Documentation**
  * Updated internal component tracking notes and size estimates across Studio panels and editors.
  * Clarified references for deferred component refactoring work.

* **Maintenance**
  * No changes to functionality, runtime behavior, public interfaces, or the user experience.

### DIFF
--- a/packages/studio/src/features/auth/organization-detail.tsx
+++ b/packages/studio/src/features/auth/organization-detail.tsx
@@ -115,7 +115,7 @@ const ManagedTable = ({
  * the whole surface also polls via {@link useAutoRefresh} since the auth store
  * is HTTP-only with no live channel.
  */
-// react-doctor-disable-next-line react-doctor/no-giant-component -- ~551 lines. Decomposing this is a real refactor with its own review, not a lint fix — deferred deliberately, and recorded under "Deferred" in plans/README.md's Wave 15 so it is not invisible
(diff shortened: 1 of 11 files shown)


prisis:
Superseded — decomposing the components outright, which deletes these suppressions rather than annotating them. Tracking issue #230 stays as the record until that lands.

PR #299, opened 2026-08-02
fix(shard-engine): encode bigint/bytes doc columns so they round-trip on the DO row store

prisis:

## Commits

- fix(shard-engine): encode bigint/bytes doc columns so they round-trip on the DO row store
- fix(sql-store): decode BLOB columns back to ArrayBuffer for v.bytes()

```
packages/shard-engine/__tests__/ctx-db.bigint-bytes.test.ts
packages/shard-engine/__tests__/ctx-db.cdc.test.ts
packages/shard-engine/src/ctx-db-cdc.ts
packages/shard-engine/src/ctx-db.ts
packages/shard-engine/src/do-sql.ts
packages/sql-store/__tests__/value-codec.test.ts
packages/sql-store/src/value-codec.ts
```


prisis:

PR #301, opened 2026-08-02
fix(sql-store): allow filter() on the bare global reader so RLS can stage a search

prisis:

## Commits

- fix(sql-store): allow filter() on the bare global reader so RLS can stage a search
- fix(server): restore async iteration on masked readers
- fix(shard-engine): fail unserializable writes distinctly instead of charging the byte cap

```
packages/server/__tests__/mask.test.ts
packages/server/src/mask/middleware.ts
packages/shard-engine/__tests__/reactive-cache.test.ts
packages/shard-engine/__tests__/transaction-headroom.test.ts
packages/shard-engine/src/estimate-bytes.ts
packages/shard-engine/src/reactive-cache.ts
packages/shard-engine/src/transaction-headroom.ts
packages/sql-store/__tests__/ctx-db-search.test.ts
packages/sql-store/src/ctx-db.ts
```


prisis:

PR #313, opened 2026-08-02
fix(codegen): reject non-identifier table names with a pinpointed diagnostic

prisis:

## Commits

- fix(codegen): reject non-identifier table names with a pinpointed diagnostic
- fix(codegen): derive the umbrella qualifier rewrite from one shared package list

```
packages/codegen/__tests__/discover-schema.test.ts
packages/codegen/__tests__/run-codegen.test.ts
packages/codegen/src/discover-schema.ts
packages/codegen/src/emit.ts
```


prisis:

PR #319, opened 2026-08-02
fix(client): promote a follower after the leader yields leadership

prisis:
Work from an agent worktree, pushed under a descriptive name (local branch was `worktree-agent-a296f70ca717ffc10`).

## Commits

- fix(client): promote a follower after the leader yields leadership
- fix(client): mirror leader connection status so follower tabs queue offline writes
- fix(client): scope the cross-tab settled watermark to the owning clientId
- fix(client,do,db): carry lastMutationId on list data frames and consume it
- fix(client): scope the cross-tab channel to deployment and identity
- fix(client): stamp and verify leader identity on cross-tab data frames
- test(client): close every fake/real-timer LunoraClient in socket-lifecycle tests
- fix(db): stop a zero mutation watermark from disabling the RPC-ack fallback
- fix(client): promote a leader tab's fresh coordinator immediately
- chore(api-snapshots): sync client and shard-engine snapshots

### DIFF
--- a/api-snapshots/client.api.md
+++ b/api-snapshots/client.api.md
@@ -1349,12 +1349,14 @@ class TabCoordinator {
     start(): void;
     stop(): void;
     isLeader(): boolean;
+    promoteImmediately(): void;
     get leaderTabId(): string | undefined;
     get id(): string;
     get isRunning(): boolean;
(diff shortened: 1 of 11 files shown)


prisis:

PR #321, opened 2026-08-02
fix(platform): rate node scheduler and alarms unsupported — jobs are never dispatched

prisis:
Work from an agent worktree, pushed under a descriptive name (local branch was `worktree-agent-a3e2c22e70fd79ddf`).

## Commits

- fix(platform): rate node scheduler and alarms unsupported — jobs are never dispatched
- fix(platform-node): make close() terminal and serialize transaction
- test(platform): give the conformance TCK teeth — dispatch, real skips, safe cleanup


prisis:

PR #323, opened 2026-08-02
fix(auth-ui): guard AuthView route lookup against prototype segments

prisis:
Work from an agent worktree, pushed under a descriptive name (local branch was `worktree-agent-adfed38045ccd5c2b`).

## Commits

- fix(auth-ui): guard AuthView route lookup against prototype segments
- fix(auth-ui): sniff empty-MIME org-logo uploads like the avatar path
- test(auth-ui): fix flaky assertion count in avatar upload test
- fix(auth-ui): gate sign-up card, route, and footer link on the discovered signUp flag
- fix(auth-ui): merge the invitation bounce params into redirects.signIn's existing query
- docs(auth-ui): record the discovery-rebuild vs in-flight form state trade-off
- fix(auth-ui): re-sync the auth-ui registry payloads from source
- ci: run the auth-ui registry drift check unconditionally

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -240,12 +240,6 @@
             - "name": "Lint (prettier)"
               "run": "pnpm run lint:prettier"
 
-            # The auth-ui-* registry items are generated from @lunora/auth-ui's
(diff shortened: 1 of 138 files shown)


prisis:

PR #329, opened 2026-08-02
fix(do): repin @lunora/platform-cloudflare so it links to the workspace

prisis:
## The bug

`packages/do` pinned `@lunora/platform-cloudflare` at `1.0.0-alpha.2`; that package is now at `1.0.0-alpha.3`. pnpm links a workspace package only when the local version **satisfies** the specifier, so once the sibling was released past the pin, resolution fell back to the **registry** — `@lunora/do` builds against a published copy of the host adapters instead of the one in this tree.

## Why CI never caught it

The committed lockfile still records `link:../platform-cloudflare`, because it was generated while the versions matched. `--frozen-lockfile` installs are therefore unaffected and CI is green.

It bites on **regeneration** — which `AGENTS.md` mandates after every merge ("discard any conflicted lockfile and regenerate it"). So the drift surfaces precisely when someone follows the documented procedure. That is how I hit it.

## Scope: deliberately one line

I first measured this as 74 stale sibling pins and was wrong about the impact. The other 73 are inert: their lockfile entries record `specifier: workspace:*`, so pnpm links them regardless of what the manifest pin says, and `pnpm install --frozen-lockfile` passes on unmodified `alpha`. Only **four** lockfile entries carry an exact specifier, and this is the one that has gone stale.

Normalizing all 74 would have been churn that fixes nothing and drifts again on the next release.

## Verification

`--frozen-lockfile` install passes · `@lunora/do` 519/519 tests, 0 type errors · `lint:package-json` clean · lockfile diff is the one specifier line.

## Spotted, not fixed here

`cron-parser@5.6.2` — which `@lunora/scheduler` pins — was deprecated upstream: *"Published tarball was built from a stale dist/ and is missing the fixes from #414 and #415 — its contents are effectively v5.6.1. Upgrade to v5.7.0."* `5.7.0` exists but is currently blocked by this repo's `minimumReleaseAge: 1440` supply-chain guard, so the bump is a follow-up once it matures.

### DIFF
--- a/packages/do/package.json
+++ b/packages/do/package.json
@@ -64,7 +64,7 @@
         "@lunora/errors": "1.0.0-alpha.11",
         "@lunora/observability": "1.0.0-alpha.4",
         "@lunora/platform": "1.0.0-alpha.2",
-        "@lunora/platform-cloudflare": "1.0.0-alpha.2",
+        "@lunora/platform-cloudflare": "1.0.0-alpha.3",
         "@lunora/shard-engine": "1.0.0-alpha.5",
         "drizzle-orm": "catalog:drizzle"
     },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2809,7 +2809,7 @@ importers:
         specifier: workspace:*
         version: link:../platform
       '@lunora/platform-cloudflare':
-        specifier: 1.0.0-alpha.2
+        specifier: 1.0.0-alpha.3
         version: link:../platform-cloudflare
       '@lunora/shard-engine':
         specifier: workspace:*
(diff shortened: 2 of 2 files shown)


prisis:
Superseded by the release on `alpha`.

`810848931 chore(deps): sync pnpm-lock.yaml with released manifests` already moved `packages/do` to `@lunora/platform-cloudflare: 1.0.0-alpha.3`, which is exactly what this PR changed. Rebasing onto the new `alpha` leaves an empty commit.

Verified on `origin/alpha`: all four exact-specifier lockfile entries are in sync, zero stale.

That also settles the open question about whether this needed a repo guard — it doesn't. The drift is a window between a sibling's release and its dependents', and `multi-semantic-release` closes it on the next release. A hard CI check would fail for the whole window, which is why one shouldn't exist. Worth knowing rather than re-investigating later.

The `cron-parser@5.6.2` deprecation noted in the description is unaffected and still stands as a follow-up — `5.7.0` remains blocked by `minimumReleaseAge: 1440` until it matures.

PR #330, opened 2026-08-02
feat(docs): add cookie consent + cookieless analytics

prisis:
## Summary

Adds a GDPR consent layer to the docs site and wires PostHog behind it in cookieless mode.

- **Consent banner + preferences dialog** (c15t v2, offline mode) rendered in the site's dark theme, with a first-layer link to `/privacy` and vendor branding hidden.
- **Jurisdiction-aware:** the root route loader reads the CDN geo headers server-side (`x-country` on Netlify, plus the Cloudflare/Vercel/CloudFront names) via a server function; the result is dehydrated to the client so both sides initialize the consent store with the same country and c15t decides whether the banner is required. Headers are advisory only — absent or spoofed headers fail safe (banner shown, GDPR assumed).
- **Withdrawal is as easy as consent:** a "Cookie settings" entry in the footer's Legal column reopens the dialog at any time.
- **Cookieless PostHog** (`cookieless_mode: "on_reject"`): before a choice and after "Reject", nothing is stored on the device — users are counted via PostHog's server-side daily-salted hash. "Accept" switches to normal persistence with full session continuity. Init is gated on `VITE_POSTHOG_API_KEY` (host defaults to the EU endpoint) and disabled in dev.

## Deployment requirements

- Set `VITE_POSTHOG_API_KEY` (and optionally `VITE_POSTHOG_HOST`) in the Netlify build environment.
- Enable **"Cookieless server hash mode"** in the PostHog project settings (Project Settings → Web analytics) — without it, cookieless events are silently dropped.

## Verification

- Type-check, ESLint, Prettier, package.json key-order gate, docs tests, and a full production `vite build` (incl. prerender) pass.
- Browser-verified: banner on first visit, Reject All persists, footer "Cookie settings" reopens the dialog, privacy link renders, no vendor branding.
- Geo bridge verified end-to-end: a request with `x-country: US` dehydrates `{country:"US"}` into the router payload that seeds the client consent store.

https://claude.ai/code/session_01EELf27TTEX6ohpmP2vPrMa

## Summary by CodeRabbit

* **New Features**
  * Added cookie consent banners and settings controls to the documentation site.
  * Added regional consent handling based on visitor location.
  * Added privacy-conscious analytics that respects measurement consent.
  * Added a cookie settings link to the site footer.
  * Added styling support for the consent management interface.

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -27,6 +27,7 @@
     },
     "dependencies": {
         "@babel/core": "catalog:web",
+        "@c15t/react": "catalog:web",
         "@fuma-comment/react": "catalog:web",
         "@fumadocs/mdx-remote": "catalog:web",
         "@icons-pack/react-simple-icons": "catalog:web",
--- a/apps/docs/src/components/sections/footer.tsx
+++ b/apps/docs/src/components/sections/footer.tsx
@@ -1,3 +1,4 @@
+import { ConsentDialogLink } from "@c15t/react";
 import DiscordLogoIcon from "@icons-pack/react-simple-icons/icons/SiDiscord.mjs";
 import GitHubLogoIcon from "@icons-pack/react-simple-icons/icons/SiGithub.mjs";
 import { Link } from "@tanstack/react-router";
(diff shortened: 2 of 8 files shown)


prisis:
Addressed all four review findings in 0ca69cc2f:

- **Placeholder country codes**: geo header values are now validated against `^[A-Z]{2}$` and the CDN placeholders `XX`/`T1` are rejected, so an unknown/Tor country can no longer read as "non-GDPR" and suppress the banner.
- **Consent withdrawal**: when measurement consent flips from granted to denied, the provider now calls `posthog.reset(true)` (guarded by `has_opted_in_capturing()`) before opting out, erasing the persisted `distinct_id`/`device_id` from the accept path.
- **Duplicate overrides**: `store: { overrides }` removed — verified in the c15t 2.1.0 bundle that jurisdiction detection reads the top-level `options.overrides`.
- **CSS import position**: the c15t styles `@import` moved into the import block above `@plugin`; confirmed the rules still land in the built CSS bundle.

Also picked up the nitpick: the root loader result is now cached for the session (`staleTime: Infinity`).

PR #334, opened 2026-08-02
fix(platform): pin error identity in the host contract

prisis:
Review follow-ups to #332. These were pushed to that branch before it merged, but the merge took the first three commits only, so they did not land.

#332 fixed the behaviour; this fixes the *assurance* and the naming.

## The property was pinned only by doubles that assumed it

`cloudflare-host.transaction.test.ts` asserts against hand-written doubles that implement the very workerd behaviours being claimed — flatten-on-rethrow and abort-on-rejection. A double cannot fail if the platform differs from the author's model of it, and it says nothing about any other host. `ShardHost` itself was silent on what happens to a thrown error, so a new `platform-<target>` could regress it and stay green.

The contract now states it on both `runSerialized` and `transaction`, and the conformance suite asserts it: a sentinel carrying `code`/`status` must come back **by identity**, and the host must still serve the next caller afterwards.

Verified under real workerd through `@lunora/do`'s `runInDurableObject` project — the new case **fails against the pre-#332 adapter and passes against the fixed one**, which is the assurance the doubles could not give. The in-memory reference host and `@lunora/platform-node` both already satisfy it unchanged.

## Also

- **Drop the explicit type argument on `blockConcurrencyWhile`.** It was discarding the closure's return type and then forcing an `as never` to recover it; `blockConcurrencyWhile<T>(cb: () => Promise<T>): Promise<T>` infers fine (`@cloudflare/workers-types@5.20260724.1/index.d.ts:675`). `as const` on the discriminant removes both.
- **Attach a failed commit/rollback as `cause`.** When the handler's error takes precedence, a broken storage layer was being discarded entirely — the opposite of the "the platform's own failure is surfaced unchanged" rule stated three lines above it.
- **Rename `symbolDeclaredUnreachable` → `declaredInHandlerModule`.** After #332 the predicate returns `true` for exported declarations, which *are* reachable; the name had become wrong and most of its docblock existed to apologise for that.
- **Cut the duplicated incident narrative.** The same ~90-word story appeared in three places. Kept the platform behaviour that cannot be derived from the code, dropped the past-tense bug story — that belongs in the commit, not in a comment that will outlive it.

## Checks

`platform` 45, `platform-cloudflare` 31, `codegen` 1045, `do` 519 plus 56 under workerd, `platform-node` 52. `api:check` clean — no public surface change.

https://claude.ai/code/session_01CYvgKochPxnrCCtKB1tXGf

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the exact original error when serialized operations or transactions fail.
  * Maintained host availability for subsequent operations after a failed transaction or serialized callback.
  * Improved handling when both an operation and its storage commit or rollback encounter errors.

* **Tests**
  * Added coverage for error preservation, rollback behavior, and continued operation after failures.

* **Documentation**
  * Clarified error-handling and recovery behavior for transactions and serialized operations.

### DIFF
--- a/packages/codegen/src/discover-functions.ts
+++ b/packages/codegen/src/discover-functions.ts
@@ -343,19 +343,15 @@ const argsFromCall = (call: CallExpression): Record<string, ValidatorIR> => {
 };
 
 /**
- * True when a type's own symbol resolves to an interface/type-alias declared in
- * the handler's source file — i.e. a name unreachable from `_generated/api.ts`.
+ * True when the type's symbol is declared in the handler's own source file.
  *
- * Exported declarations count too. The checker prints an exported local type by
- * name (it is nameable from the handler), but codegen only ever emits an import
- * for a qualifier the checker itself rendered as `import("…")` — it never
- * synthesises one for the handler's own module. So `export interface Board` came
- * out as a bare `Board` in `_generated/api.ts` with nothing importing it: TS2304
- * in generated code, while `lunora codegen` exits 0. Structurally expanding it,
- * exactly as a non-exported declaration already was, keeps the emitted type
- * identical in shape and resolvable from anywhere.
+ * `_generated/api.ts` never imports that module — codegen only emits an import
+ * for a qualifier the checker itself rendered as `import("…")` — so such a name
+ * must be expanded structurally rather than printed. Exported declarations are
+ * included: being nameable from the handler is what makes the checker print the
+ * bare name, which is exactly the case that does not resolve from `_generated/`.
  */
-const symbolDeclaredUnreachable = (type: Type, handlerFilePath: string): boolean => {
+const declaredInHandlerModule = (type: Type, handlerFilePath: string): boolean => {
     for (const candidate of [type.getSymbol(), type.getAliasSymbol()]) {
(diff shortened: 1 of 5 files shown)


prisis:
Fixed in `a56ddd421` — thanks, this was a real one.

I reproduced it before accepting: `Object.freeze(new Error(...))` then `err.cause = other` throws `TypeError` under strict mode, so the assignment would have propagated **in place of** the handler's error. That is exactly the loss this wrapper exists to prevent — the line meant to preserve information could destroy it.

Taken with two changes to the proposed diff:

1. **Extracted into a named `attachCause` helper.** The inline nested `try` pushed `transaction` to cognitive complexity 18 against a limit of 15 (`sonarjs/cognitive-complexity`), so the extraction was needed regardless — and it reads better than the nesting.
2. **`Object.defineProperty` rather than assignment**, which states the intent (add an own property to this instance) and does not trip `no-param-reassign`.

Covered both directions: a platform failure is attached as `cause` when the error can carry one, and a frozen error still arrives by identity when it cannot. I verified the second test fails without the guard, and re-ran the conformance leg under real workerd (56 passed).

PR #338, opened 2026-08-03
fix(shard-engine,sql-store): round-trip bigint and bytes columns on the DO row store

prisis:
Two related data-integrity bugs on the Durable Object row store, both hit by
schemas that use `v.bigint()` or `v.bytes()` columns.

## The defects

**1. Documents were serialized with a bare `JSON.stringify` / `JSON.parse`.**

- a `v.bigint()` column threw a raw `TypeError` out of `ctx.db.insert`
- a `v.bytes()` column **silently corrupted to `{}` on write while reporting
  success** — the write returned normally and the data was gone

**2. `sqliteDecode` had no `"bytes"` case.**

A decoded BLOB came back as whatever the driver returned verbatim —
`ArrayBuffer` on workerd D1, `Uint8Array` on `node:sqlite`, `Buffer` on
pg/mysql2 — while `v.bytes()` validates `value instanceof ArrayBuffer`. So
re-validation failed on every backend whose driver doesn't hand back a genuine
`ArrayBuffer`.

## The fix

Route every doc-blob write/read — `insert`, `insertManyUnsafe`, `patch`,
`replace`, soft delete, the OCC CAS snapshot, and the CDC changelog — through
`encodeDocJson` / `decodeDocJson` in `do-sql.ts`, wrapping the existing wire
codec so both types round-trip.

Add a `"bytes"` branch to `sqliteDecode` that normalizes any driver return
shape to a right-sized `ArrayBuffer`, slicing on the view's own `byteOffset` /
`byteLength` so a Buffer-pool view cannot leak neighbouring bytes.

## Compatibility

A document with no bigint/bytes/Date/Map/Set leaves still encodes
byte-identically to plain `JSON.stringify`, so existing rows keep parsing
unchanged and the OCC compare-and-swap still matches.

## Verification

- `@lunora/shard-engine` — 978 pass (new `ctx-db.bigint-bytes.test.ts`, 10 cases)
- `@lunora/sql-store` — 104 pass (new bytes cases in `value-codec.test.ts`)
- `lint:types`, `eslint`, `api:check` 44/44, `lint:package-json` 74/74

Ablation check: reverting the `case "bytes"` block fails 2 tests, including the
Buffer-pool leak case. The shard-engine suite alone does **not** catch it — it
covers the encode side only, so both test additions are load-bearing.

https://claude.ai/code/session_019WfuWNN3cWJE6beZjng347

## Summary by CodeRabbit

* **New Features**
  * Added support for storing and retrieving documents containing large integer values and binary data.
  * Improved handling of byte arrays and typed binary values while preserving their visible contents.

* **Bug Fixes**
  * Improved document encoding and decoding across database operations, change tracking, inserts, updates, replacements, and deletes.
  * Maintained compatibility with existing plain JSON documents and ensured reliable handling of binary data ranges.

### DIFF
--- a/packages/hyperdrive/__tests__/global-mysql.integration.test.ts
+++ b/packages/hyperdrive/__tests__/global-mysql.integration.test.ts
@@ -197,7 +197,7 @@ describe("hyperdrive global — MySQL (mysql-memory-server) integration", () => 
         it(
             "round-trips bigint, bytes, boolean, object, number and string through MySQL",
             async () => {
-                expect.assertions(6);
+                expect.assertions(7);
 
                 await runSqlGlobalTableMigrations(harness.exec, typesSchema, mysqlDialect);
                 const writer = writerFor(typesSchema);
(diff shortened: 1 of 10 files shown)


prisis:
Fixed in 6a25ca51a — the finding reproduced, so it was a real gap.

```
current impl -> SharedArrayBuffer | instanceof ArrayBuffer: false
proposed fix -> ArrayBuffer       | instanceof ArrayBuffer: true
```

Switched the view path to `new Uint8Array(raw.buffer, raw.byteOffset, raw.byteLength).slice().buffer`. That still narrows to the view's own window, so the Buffer-pool leak stays closed, but always yields a genuine `ArrayBuffer` regardless of backing store.

Added a regression test for the shared-memory case and ablated it: with the old `raw.buffer.slice(...)` restored the new test fails, with the fix all 22 pass.

PR #354, opened 2026-08-06
feat(import): migrate Convex `_storage` blobs to R2

prisis:
## Summary

`lunora import` moved a Convex export's **documents** but not its **files**. The
`_storage` table was skipped with a warning telling the operator to "upload the
exported blobs to R2 and re-point the keys" by hand — so a Convex app with file
storage lost every file, and every field holding a storage id became a dangling
reference. This closes that gap and adds the verification the command never had.

```bash
npx convex export --path ./convex-export --include-file-storage

lunora import ./convex-export --scan            # propose the column mapping
lunora import ./convex-export --with-storage --verify
```

**`--with-storage`** uploads every blob under a content-hash key
(`sha256` hex of the bytes) through a checksum-verified admin route, then
rewrites the documents' references to the new keys. Blobs land before documents,
so no imported row can reference an object that is not there yet. Re-runs are
cheap and safe: one prefix listing tells the CLI what is already present at the
right size, and those are mapped without re-uploading.

Reference rewriting distinguishes what it can know from what it cannot.
`{ $storage: id }` is Convex's self-describing form, so it is rewritten wherever
it occurs at any depth. A plain string is ambiguous against ordinary text, so it
is rewritten only under a column `lunora/import-convex.json` names — which is
what **`--scan`** generates, by exact-matching source values against the
`_storage` id set. It writes the file with `wx`, so a reviewed mapping is never
clobbered by a re-scan, and it imports nothing.

**`--verify`** checks that every source row is accounted for and that every
storage reference resolved. The two failure modes are kept apart because they
have different remedies: a reference whose blob does not exist is broken data no
mapping can fix, and fails the run; a string matching a migrated blob in an
unnamed column may equally be user text that happens to look like an id, so it
warns and names the column to add.

Directory exports and `snapshot.zip` are both accepted.

### Also in here

- **Runtime** (`@lunora/runtime`) — `PUT /_lunora/admin/storage` accepts
  `expectedSha256`/`expectedSize`, digests the body, and rejects with
  `STORAGE_CHECKSUM_MISMATCH` **before** anything is written. The route declares
  a 32 MiB body budget of its own the way the KV value PUT already does; without
  it the shared 1 MiB JSON cap would push nearly every photo onto the fallback
  path. `/_lunora/admin/storage/url` now validates `method` against a GET/PUT
  allowlist rather than casting it — an unchecked `?method=DELETE` would have
  minted a signed destructive URL.
- **Codegen** (`@lunora/codegen`) — the generated `storageSignedUrl` adapter
  declared only `{ bucket, expiresInSeconds }` and dropped the `method` and
  `contentType` that both the runtime contract and `@lunora/storage` accept.
  `getSignedUrl` then defaulted to `GET`, and since the method is baked into the
  HMAC canonical, any caller asking for a presigned PUT received a URL that
  verifies as a GET. This is the first caller to ask for one.
- **Structure** — `data-transfer.ts` had grown to ~1,650 lines holding seven
  concerns. It is now a directory of focused modules, matching how every other
  multi-file command in `@lunora/cli` is laid out, and `runImportCommand` is a
  sequence of named phases rather than one long function. Every
  `sonarjs/cognitive-complexity` suppression in the touched files is gone — they
  were removed by extraction, not by widening the threshold.
- **Plans** — `plans/304` records the answers to its open questions (content-hash
  keys, skip-if-present idempotency without a `--force`, the cut FK spot check
  and why, the 32 MiB cap). `plans/305` gains a per-asset coverage table so the
  Supabase/Firebase work states up front what "migrated" does and does not
  include.

### Known limitation

Objects above 32 MiB cannot reach the worker, so they fall back to a signed
`PUT`. That path needs the app to have signed URLs configured **and** to serve
the route those URLs address — Lunora does not mount one — and it is verified
only after the write (by size, plus SHA-256 where the bucket records one, which
R2's `list()` does not). An object that fails verification is deleted rather than
left to look already-migrated. `from-convex.mdx` says all of this, and points at
`wrangler r2 object put` for anyone without a signed-PUT route.

## Linked issues

- Refs `plans/304-convex-blob-migration-and-import-verification.md` (W1–W5)
- Unblocks `plans/305-supabase-firebase-importers.md` (its storage transfer
  reuses the verified upload route)

## Test plan

- [x] `pnpm --filter "@lunora/cli" run test` — 1111 passing (33 new, covering
      blob upload, both reference shapes, nested rewriting, base64 vs hex
      digests, idempotent re-runs, `keyPrefix`, ZIP snapshots, the signed-PUT
      path including delete-on-mismatch, cursor paging, byte-ceiling batching,
      partial-failure reporting, and every flag guard)
- [x] `pnpm --filter "@lunora/runtime" run test` — 844 passing (checksum
      mismatch rejects before write, an upload above the shared JSON cap
      succeeds, the signed-URL method allowlist)
- [x] `pnpm --filter "@lunora/codegen" run test` — 1054 passing, golden fixtures
      and example `_generated/app.ts` regenerated
- [x] `pnpm --filter "…" run lint:types` clean for cli / runtime / codegen
- [x] `eslint` and `prettier --check` clean on every touched file
- [x] `pnpm run lint:package-json` — 74/74 sorted
- [x] Built `dist/` smoke-tested end-to-end through the ZIP path, to confirm
      `adm-zip`'s CJS interop survives bundling into the ESM-only CLI

## Checklist

- [x] Updated relevant docs in `apps/docs` (`migrating/from-convex.mdx`)


prisis:
@coderabbitai review

prisis:
@coderabbitai review

PR #361, opened 2026-08-08
feat(cli): make the automation surface machine-readable

prisis:
Four related changes to the surface an automated consumer — a CI job, a coding agent, a deploy pipeline — actually talks to. The capabilities were mostly already here; the machine-readable edges were inconsistent.

## What changes

**`lunora doctor` becomes consumable.** It was the only preflight command with `options: []` and findings identified solely by an English sentence. Now `--format json` with an `{ ok, code, summary, findings }` envelope, a required `code` on every finding drawn from a 17-entry literal union (a typo fails `lint:types`), and a new `cli-shadowed` check for a global binary shadowing the project's. The docs table *is* the fixture — a test parses the code column out of `docs/index.mdx` and asserts it equals `DOCTOR_CODES`, so adding a code without documenting it fails the suite.

**`lunora deploy` reports what it deployed.** It already parsed the deployed URL out of wrangler's output, but the gate turned that off for `--format json` and `--preview` — the two paths automation uses. `DeployCommandResult` now carries `deployment` (`url`, `workerName`, `env`, `dryRun`, `preview`, `deployedAt`); json mode captures via `captureStdoutSilently` and replays wrangler's log to stderr so stdout stays exactly one document. Links refresh on every real deploy: write when absent, no-op when equal, warn-and-keep when different — a stale link silently misrouting `--migrate` at a decommissioned URL is worse than the overwrite it avoids. New `--health-check` probes `/_lunora/health/ready` (falling back to the aggregate) with a bounded retry before `--migrate` runs; `verify` now shares that probe.

**MCP gains the reads an agent needs to debug.** It could call a deployment but not see what happened. Five token-gated read tools — `lunora_get_logs`, `_get_issues`, `_get_advisories`, `_get_query_insights`, `_get_migration_status` — plus `outputSchema`/`structuredContent` alongside the retained text block. These are read-only but *privileged*, so they are omitted from `ListTools` without an admin token and refused at dispatch, independent of `--allow-writes`.

**The deployed Worker gets weighed.** Nothing measured it. `lunora build` now reports raw + gzip size (`gzipSync` reproduces wrangler's own `Total Upload` line to the byte), and a `worker-size` CI job gates a reference template against a committed baseline. It is its own job, deliberately not in `postinstall`, where one failure turns every job red in its setup step.

## BREAKING CHANGE

`lunora build --format json` now emits its own result document instead of deploy's. There was nowhere to put the size field otherwise. Consumers reading build's json output see a different shape.

## Notable findings

**35% of a hello-world Worker is an English NLP library.** `@visulima/redact` hard-depends on `compromise@^14.15.1` — 606 KiB raw in the bundle, larger than any first-party package — reached from `packages/observability/src/request-log.ts:30`. Every app carries it. Not fixed here; recorded as a follow-up.

A reference worker measures **412.9 KiB gzipped against a 3 MB Free / 10 MB Paid ceiling** — 13.4%. That is why only the CI gate shipped and the planned user-facing warning was dropped: it would have been an alarm nobody could legitimately trip, at a threshold nobody could justify.

Three of the four plans were wrong about something, corrected in-flight and recorded in `plans/README.md`. The one worth repeating: the planned pnpm shadow-detection would have warned on every pnpm project, because pnpm writes a shell shim where the design assumed a symlink.

## Verification

Run against the merged tree, not the individual branches:

```
CLI        88 files / 1203 tests passed
MCP        11 files /  181 tests passed
api:check  47/47 snapshots
lint:types clean (cli + mcp)
```

`dist:check`'s `jsxDEV` findings reproduce on untouched `alpha` — a dev-vs-prod build artifact, not from this branch.

https://claude.ai/code/session_01KBeSX2o4sTCPjVDRDWkVQG

## Summary by CodeRabbit

* **New Features**
  * Added optional post-deployment health checks with retries and readiness-route fallback.
  * Added structured JSON output for `lunora doctor` and `lunora deploy`.
  * Added stable diagnostic codes and clearer environment diagnostics.
  * Added read-only observability tools for logs, issues, advisories, query insights, and migration status, protected by admin tokens.
  * Improved deployment URL reporting and automatic project linking.
  * Added automated Worker bundle size checks against compressed and raw size limits.

* **Documentation**
  * Documented Worker size limits, health checks, JSON output, deployment linking, and observability tools.


prisis:
All six review comments addressed in d5ac5fb3a, plus one thing the `alpha` merge surfaced.

**`health-probe.ts` — per-attempt timeout.** Correct and the most serious of the six. Node's `fetch` has no timeout, so a worker that accepts the connection and then goes quiet would have hung `deploy --health-check` until CI's job timeout killed it — which defeats the fixed attempt budget the probe exists to provide. Each attempt now carries `AbortSignal.timeout` (default 10s, `timeoutMs` overridable); an injected `fetchImpl` still owns its own deadline. Covered by a test that asserts the signal reaches `fetch`, rather than waiting on a real timeout.

**`file-filters.yml` — the gate could be switched off by editing the gate.** Correct. Added a `worker_size` filter covering the checker, the baseline, and the workflow. One thing the comment did not mention and would have made the fix a no-op: `files-changed` declares its outputs explicitly, so `worker_size` also had to be added there — otherwise the new `if` clause reads an always-empty output and never fires. Wired all three.

**64 MB pre-compression limit.** Correct, and I verified it against the Cloudflare limits page rather than taking it on trust: 3 MB Free / 10 MB Paid after gzip, and 64 MB before compression on both plans. The gzip check is a *baseline* comparison, so it cannot see a bundle that compresses unusually well — a large generated table is tiny gzipped and enormous raw, passes the gate, and is rejected at upload. Raw now gets its own absolute check with a message saying `worker-size:update` will not help, since it is a platform limit rather than a budget. Docs updated to state both limits and to stop telling readers gzip is the only figure to compare.

**Doctor JSON example, CI snippet, MCP README.** All three fixed as described — `jsonc` with a truncation comment, `set -euo pipefail` plus a URL guard, and the `lunora_get_migration_status` exception documented.

**Separately:** merging `alpha` down showed `@lunora/mcp` was the only one of ten consumers still pinning `@lunora/shard-engine@1.0.0-alpha.15` while the release moved everything else to `alpha.16`. The `overrides:` block forces `workspace:*` locally so it is invisible in the repo, but it is what the published manifest would carry. Fixed in bf34e65fd. Lockfile regenerated after the merge rather than text-merged; it came back with zero drift.

Verification on the merged tree: CLI 1204/1204, MCP 181/181, `lint:types` clean, prettier clean, `lint:package-json` 74/74 sorted.

prisis:
Correction to my previous comment: it claimed all six review comments were addressed, which was true of the six *actionable* ones but not of everything outstanding. Three more were open, now fixed in a2f54ddb6.

**`.github/file-filters.yml` not listed in its own filter** (the 7th inline comment, from the re-review after my push — I had not re-fetched). Correct, and it is the same hole as the one the filter was added to close, one level up: a PR editing the filter definition would not match the filter, the gate would skip, and `test-required-check` passes on skipped jobs. Added the file to its own filter.

**`verify` rebuilt the health URL** instead of reading `probe.url`. The two agree only while `verify` keeps the default `paths`; the moment it passes more, the success message names a URL the verdict did not come from. Fixed, `joinHealthUrl` import dropped.

**Loose list item** in the doctor-checks list — stray blank line removed.

CLI 1204/1204, `lint:types` clean, prettier clean.

PR #387, opened 2026-08-09
feat(runtime): record what backup retention deleted

prisis:
The scheduled backup's retention deletes objects from R2 and, until now, said nothing when it worked. The only log was on the failure path.

That is the gap behind plan 313 §4.4 ("No silent deletion of anything"). It is narrower than it sounds — retention is **opt-in**, so with no `backupRetain` configured nothing is ever deleted — but when an operator does configure it, snapshots were disappearing with no record.

**Both retention defects found during review were silent successes**: the cron deleting an operator's hand-taken snapshots, and later deleting another deployment's. Each was reconstructed afterwards from a missing file. A line naming what was pruned turns the next one into something you notice.

## What it does

Names the retention window and the deleted keys, capped at 10 with a `(+N more)` summary — a first run after enabling retention on an old bucket can delete many, and a log line nobody can read is its own kind of silence. Stays quiet when it deletes nothing, so an ordinary cron fire adds no noise.

It uses the same `console` channel as the existing retention-failure warning; a Worker has no other operator-visible sink at this point in the run.

## What it deliberately is not

A forensic record, not an approval gate. R2 deletes are irreversible and this is emitted *after* the fact — an operator who wants to review a prune before it destroys anything wants the `prune` verb with `--dry-run`, which is plan 313's WS4 and still open. The comment says so, so the next reader does not mistake this for the fix.

## Verification

`@lunora/runtime` **857 tests** (one new). The new test asserts both halves — that a prune names what it deleted, and that a no-op prune logs nothing — and I confirmed it is not vacuous by removing the log and watching only that test fail (`expected \"info\" to be called`), then restoring.

`lint:types` and `lint:eslint --max-warnings=0` clean, Prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduled backup retention handling when no backups are stale.
  * Retention cleanup now processes snapshot and related metadata deletions more efficiently.
  * Deletion logs now show up to 10 removed backup keys and report any additional deletions by count.


prisis:
Superseded by #389, merged as `01dc8bf40`.

This branch is an ancestor of `feat/backup-prune-verb`, so its commits went in with that merge. Because the repo squash-merges, they do not appear in `alpha`'s history under their own shas and this PR will keep reporting as unmerged — the content is in.

Verified on `alpha` rather than inferred from PR state: the retention preview route, both retention log records, the `retention` and `prune` CLI verbs, and `pruneBackups` being reachable only from `runBackupPrune` (the scheduled path calls `selectStaleBackups` and never deletes).

PR #388, opened 2026-08-09
feat(runtime): preview what backup retention would delete

prisis:
## What

`lunora backup retention` — what the platform's retention would delete on its
next run, and nothing else.

```bash
$ lunora backup retention --url https://api.example.com
backup retention would keep the newest 3 of 5 under backups/ and delete 2
  backups/lunora-backup-2026-06-01T03-00-00-000Z.ndjson
  backups/lunora-backup-2026-06-02T03-00-00-000Z.ndjson
```

Backed by `GET /_lunora/admin/backup/retention`, admin-gated like the storage
read route.

## Why this and not a log

`backupRetain` deletes on every cron run, an R2 delete has no undo, and the
thing deleted is the backup. #387 records what *was* destroyed; nothing could
say what *will* be. So the only way to learn what a retention setting does on a
real bucket was to let it run and read the aftermath.

That matters more than usual here because eligibility depends on the
`lunoraBackupCron` metadata marker, which sidecars written before that feature
lack. On a pre-existing bucket, retention may own a fraction of what the bucket
holds — and no amount of reading the config tells you which fraction. The
preview does, and says so explicitly when the answer is zero.

## The surface, and why

**`lunora backup retention`** — a report verb, not `prune --dry-run`.

- A verb named `prune` that refuses to prune is confusing, and one that previews
  by default with a `--apply` behind it advertises a deleter this PR does not
  build. `retention` names what it reports on, implies nothing destructive, and
  leaves `prune` unclaimed for the actual deleter when WS4's remaining half
  lands — so that name is not spent on something that refuses.
- Not a flag on `list`, because it answers a different question from a different
  source: `list` reads the bucket, this reads the worker's own configuration.
  `list` also works against a `--dir` destination, where retention does not
  exist.
- Nothing in this PR deletes, so there is no unsafe path to default away from.

## Preview and prune are the same selection

`pruneBackups` already computed its set before deleting. That computation is now
`selectStaleBackups`, and both the prune and the preview call it — the preview
does not re-derive eligibility, because a preview that can disagree with the
prune is worse than no preview, and this subsystem has twice produced the bug
where one rule was written in two places.

The test that matters seeds **one** bucket with every case that makes
eligibility subtle — a legacy sidecar with no marker, a sidecar carrying another
deployment's cron expression, an operator-written snapshot — then runs the
preview and a real cron fire against identical bucket state and asserts the sets
are equal. Getting the inputs genuinely identical took care: the fire writes a
snapshot before it prunes, so the seed includes a snapshot at the fire's own key
and the fire overwrites it rather than adding one. Without that the two would be
answering about different buckets and an equal result would mean nothing.

**Non-vacuity checked both ways.** Perturbing `selectStaleBackups` moves both
sides together and the parity test stays green (correctly — that is what sharing
the code buys); perturbing *only* the preview fails it immediately.

## Nothing here deletes

- The route is `GET`, calls only the selection, and never reaches a `delete`.
- A test asserts the bucket is byte-for-byte unchanged after a preview — same
  keys, same contents.
- Retention itself stays on the cron. No second deleter.
- Unauthenticated → 403 before the bucket is read, so no object key can leak;
  asserted.

## Plan

Plan 313's §4.4 amendment claimed WS4 was wholly open. Narrowed: the preview
half shipped, the deletion becoming an operator-invoked step is what remains,
along with a way to run the cron backup without the prune attached.

## Note for the merge order

#387 (the retention log) is still open and touches the same function. It
refactors `pruneBackups` into "compute `stale`, then delete, then log"; this PR
moves that same computation into `selectStaleBackups`. Whichever lands second
keeps both — the log line and the shared selection are independent. The preview's
output is deliberately phrased like #387's log line, so a prediction and the
record of a real run read as one story.

### DIFF
--- a/api-snapshots/lunora.api.md
+++ b/api-snapshots/lunora.api.md
@@ -3589,6 +3589,10 @@ Re-exported from `@lunora/runtime` — signature tracked at its source.
 
 Re-exported from `@lunora/runtime` — signature tracked at its source.
 
+### `BackupRetentionPreview` (interface)
+
+Re-exported from `@lunora/runtime` — signature tracked at its source.
+
 ### `BackupStore` (interface)
 
 Re-exported from `@lunora/runtime` — signature tracked at its source.
--- a/api-snapshots/runtime.api.md
+++ b/api-snapshots/runtime.api.md
@@ -362,6 +362,18 @@ interface BackupManifestEntry {
 }
 ```
 
+### `BackupRetentionPreview` (interface)
+
+```ts
+interface BackupRetentionPreview {
+    cron?: string;
+    eligible: number;
+    keep: number;
+    prefix: string;
+    wouldDelete: string[];
+}
+```
+
 ### `BackupStore` (interface)
 
 ```ts
--- a/apps/docs/src/content/docs/concepts/backups.mdx
+++ b/apps/docs/src/content/docs/concepts/backups.mdx
@@ -149,6 +149,12 @@ way the answer is `--tables` / `backupTables`, or `--dir` plus
 
 ### Scheduled backups into R2
 
+<Callout>
+    Retention deletes, and an R2 delete has no undo. `lunora backup retention` reports what the next scheduled run would remove — from the same selection the
+    prune uses, so the preview cannot drift from the deletion — and removes nothing itself. Run it after changing `backupRetain`, and on any bucket that
+    predates the feature: snapshots written before retention learned to mark its own are never eligible, so it may own fewer of them than the bucket holds.
+</Callout>
+
 For a schedule that runs on the platform itself rather than in CI, the worker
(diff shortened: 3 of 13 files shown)


prisis:
Superseded by #389, merged as `01dc8bf40`.

This branch is an ancestor of `feat/backup-prune-verb`, so its commits went in with that merge. Because the repo squash-merges, they do not appear in `alpha`'s history under their own shas and this PR will keep reporting as unmerged — the content is in.

Verified on `alpha` rather than inferred from PR state: the retention preview route, both retention log records, the `retention` and `prune` CLI verbs, and `pruneBackups` being reachable only from `runBackupPrune` (the scheduled path calls `selectStaleBackups` and never deletes).

PR #396, opened 2026-08-10
fix(shard-engine): size emitted SQL to workerd's SQLite limits

### LINKED ISSUE #385: Bare-id `db.delete`/`db.patch` build an unscoped UNION ALL probe; fails on Workerd ("too many terms in compound SELECT")
> **Note on authorship:** This issue was written and filed by Claude, an AI assistant (Claude Code by Anthropic), on behalf of Cameron Pak ([@cameronapak](https://github.com/cameronapak)). Cam hit this bug in production, reviewed this report, and asked me to file it for him. Every claim below links to source or to the public fix PR. If anything is unclear, Cam is the human to talk to.

## Describe the bug

A bare-id `ctx.db.delete(id)` or `ctx.db.patch(id, values)` call (no `expectedTable`) resolves the id by building a `UNION ALL` SELECT across every non-global table in the schema. On Cloudflare Workers, Workerd caps compound SELECTs at 5 terms, so this probe fails at runtime once the schema has enough tables:

```
Error: too many terms in compound SELECT: SQLITE_ERROR
```

Evidence for each piece:

- **The unscoped probe (current `alpha` branch).** `nonGlobalTableNames` only filters when `expectedTable` is passed; with it `undefined`, every non-global table is included: [`packages/shard-engine/src/ctx-db.ts#L2294-L2298`](https://github.com/anolilab/lunora/blob/95d33d6703b12d1e824b3178080f8e389843f958/packages/shard-engine/src/ctx-db.ts#L2294-L2298). `locateRowById` then joins one branch per table with `UNION ALL`: [`ctx-db.ts#L2311-L2330`](https://github.com/anolilab/lunora/blob/95d33d6703b12d1e824b3178080f8e389843f958/packages/shard-engine/src/ctx-db.ts#L2311-L2330). The batch variant `locateTablesByIds` chunks by SQL-variable count (`900 / tableCount`) but not by compound-term count: [`ctx-db.ts#L2374-L2406`](https://github.com/anolilab/lunora/blob/95d33d6703b12d1e824b3178080f8e389843f958/packages/shard-engine/src/ctx-db.ts#L2374-L2406).
- **The Workerd limit.** Workerd hard-sets `SQLITE_LIMIT_COMPOUND_SELECT` to 5: [`src/workerd/util/sqlite.c++#L1382-L1384`](https://github.com/cloudflare/workerd/blob/19b1657bd4d53381011b105641e25f1e2da93fd5/src/workerd/util/sqlite.c%2B%2B#L1382-L1384). (As far as we can tell, Cloudflare's Durable Objects limits docs do not mention this one; the workerd source is the reference.)
- **`asId` does not scope.** `ctx.db.asId("nodes", id)` reads like it scopes the lookup to `nodes`, but it is compile-time branding plus structural format validation only — no runtime table scoping: [`ctx-db.ts#L2539-L2547`](https://github.com/anolilab/lunora/blob/95d33d6703b12d1e824b3178080f8e389843f958/packages/shard-engine/src/ctx-db.ts#L2539-L2547). Only the generated per-table facade (`ctx.db.nodes.delete(...)`) forwards `expectedTable` — and your own JSDoc on `nonGlobalTableNames` describes that scoping as intentional. The bare `ctx.db.delete`/`patch` signatures leave `expectedTable` optional, so nothing pushes a caller toward it, while the `asId` branding actively suggests scoping already happened.

**What we observed in production:** Dotflowy ([public repo](https://github.com/cameronapak/dotflowy)) runs a 5-table shard schema on `lunorash@1.0.0-alpha.107`. Every MCP write path that used bare `ctx.db.delete`/`ctx.db.patch` (with `asId`-branded ids) failed with the error above on Cloudflare Workers, and moving those call sites to the per-table facade fixed it. The fix PR with the failing paths and the change is public: [cameronapak/dotflowy#335](https://github.com/cameronapak/dotflowy/pull/335).

One caveat, stated to avoid overclaiming: the alpha.107 resolver we hit is not the same code as the current `alpha` branch (the current comment above `locateRowById` says the previous code probed tables in two steps), so we cannot cite the exact query alpha.107 built. What we can say from source is that the **current** resolver unions one branch per non-global table, so any schema with more than 5 non-global tables must exceed Workerd's 5-term cap on a bare-id call, and a 5-table schema sits exactly at the cap.

## Minimal reproduction

1. Define a schema with 6+ non-global tables (or 5+, to reproduce what we hit on alpha.107). Deploy to Cloudflare Workers.
2. In any mutation: `await ctx.db.delete(ctx.db.asId("someTable", existingId))`.
3. The call throws `too many terms in compound SELECT: SQLITE_ERROR`.
4. The same delete through the facade (`ctx.db.someTable.delete(id)`) succeeds.

Our real-world schema (5 tables: `nodes`, `tagColors`, `savedQueries`, `dailyIndex`, `migrateState`) is public: [`lunora/schema.ts`](https://github.com/cameronapak/dotflowy/blob/main/lunora/schema.ts).

## Suggestions (any one would help)

1. Split or chunk the bare-id probe by compound-term count (max 5 on Workerd), not only by SQL-variable count, so it works regardless of table count.
2. Or derive the table from the `Id<T>` brand at runtime (encode the table into the id at mint time), so `asId` actually scopes.
3. Or deprecate the bare-id overloads on the typed ctx (or make `expectedTable` required there) and steer users to the facade. You already test facade scoping ([`__tests__/ctx-db.by-id-table-scope.test.ts`](https://github.com/anolilab/lunora/blob/95d33d6703b12d1e824b3178080f8e389843f958/packages/shard-engine/__tests__/ctx-db.by-id-table-scope.test.ts)), so the facade is clearly the intended path; the bare overloads are the trap.

## Environment

- `lunorash` `1.0.0-alpha.107` (where the production failure occurred)
- Current `alpha` branch @ [`95d33d6`](https://github.com/anolilab/lunora/tree/95d33d6703b12d1e824b3178080f8e389843f958) (unscoped bare-id path still present, per the permalinks above)
- Cloudflare Workers / Durable Objects SQLite (Workerd)

## Affected package(s)

`@lunora/shard-engine` (via `lunorash`)

prisis:
Fixes #385, plus three more limits from the same source that the audit it prompted turned up.

## The problem

Workerd's SQLite build — the one behind **both** Durable Object storage and D1 — sets several limits far below stock SQLite. Over any of them, the statement fails at runtime with `SQLITE_ERROR`:

| limit | workerd | stock | what was over it |
| --- | --- | --- | --- |
| `COMPOUND_SELECT` | **5** | 500 | the by-id probe (#385) |
| `VARIABLE_NUMBER` | **100** | 500,000 | `IN (…)` lists, bulk `INSERT` |
| `LIKE_PATTERN_LENGTH` | **50** | 50,000 | `contains` filters |
| `EXPR_DEPTH` | **100** | 1,000 | already covered by `MAX_SEARCH_TERMS = 16` |

None of it reproduces locally: `node:sqlite` runs the stock limits and exposes no way to lower them.

## The fixes

**Compound SELECT (#385).** The bare-id probe unions one branch per non-global table, so `ctx.db.delete(id)` / `ctx.db.patch(id, …)` without an `expectedTable` threw `too many terms in compound SELECT` on any schema wider than five tables — a five-table schema sits exactly at the cap. `unionAll` nests the branches five at a time: a sub-select restarts the parser's term counter, so it stays one round-trip and returns the same rows at any branch count. The FTS vocabulary scorer, whose branches scale with query terms, goes through it too.

**Bound parameters.** `where: { id: { in: [...] } }` bound one placeholder per item, and the relation semijoin deliberately resolves up to 5,000 join keys; the batched id→table probe budgeted against 999; bulk `INSERT` bound three per row against a 500-row batch cap; the global search companion inserted 50 rows (150 parameters) per statement. `sqliteInList` renders a long list as a single `json_each(?)` parameter — same membership, placeholder count independent of set size — and keeps short lists literal so an index still plans against them. `INSERT`s chunk at 32 rows.

**LIKE patterns.** `contains` compiled to `LIKE '%' || term || '%'`, so a filter term over ~48 characters failed the query outright, as did a Studio data-browser filter or a 50-character search term on D1. It now compiles to a position function per dialect — `instr(lower(x), lower(term))` on SQLite, `LOCATE` on MySQL, `strpos` on Postgres — each folding case the way that dialect's `LIKE` did, so results are unchanged. The wildcard escaping goes away with it: a position test takes the term literally, so a client-supplied `%` can no longer become a live pattern.

## What is *not* in here

The issue also suggested encoding the table into the id so `asId` scopes at runtime, or removing the bare-id overloads. Neither is here — the probe now works at any table count, and the IDOR concern is already handled by `expectedTable` on the per-table facade. That `asId` reads like it scopes when it doesn't is worth revisiting separately.

## Testing

The caps can't be exercised locally, so `packages/shard-engine/__tests__/workerd-sql-limits.test.ts` asserts each on the **rendered** SQL — compound width per nesting depth, placeholder count, absence of a LIKE pattern — and pairs each with a behavioural case proving the reshaped statement returns the same rows on a real SQLite build. The compound-width assertions were confirmed to fail without the nesting, and the batched probe was confirmed to actually be reached.

`sqlite3` supports `.limit compound_select 5`, which is how the 5-vs-6-term boundary and the sub-select reset were verified by hand.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved support for large ID lists, complex queries, and bulk inserts without exceeding SQLite limits.
  - Prevented failures caused by long search terms and SQL pattern limits.
  - Preserved accurate document retrieval, filtering, and full-text search behavior.

- **Improvements**
  - Improved case-insensitive substring searching across supported databases.
  - Enhanced query handling for large datasets and broad schemas.
  - Improved reliability for bulk operations and large searches.


cameronapak:
Thank you!!

PR #401, opened 2026-08-10
test(templates): close the template typecheck floor

prisis:
## Why

The template smoke matrix had two holes that let real defects through.

`standalone` is the only template with no `build` script, so it was recorded `PASS` on scaffold + install alone — nothing ever compiled the code it ships. It also has no framework dependency, so the auth-ui payload gate skipped it too.

Everywhere else the gate failed only on diagnostics under `lunora/auth-ui/**` and merely *counted* each template's own errors as somebody else's problem. Nine had accumulated that way. Two were real defects, not missing declarations.

## What changed

**The build-less path now compiles.** `standalone` runs its own `codegen` script (every source file imports `#lunora/_generated/*`, which does not exist before that) and then `tsc --noEmit`.

**Two real defects, in three templates:**

- `LunoraProvider` takes `client`, not `url`. `react-router` and `tanstack-start-react` passed `url`, so the provider mounted with an undefined client and every `useQuery` below it resolved against nothing.
- Supplying a client exposed the next problem: building it at module scope shares one instance across concurrent server renders, so one visitor's cached results — and eventually their identity — leak into the next request's HTML. All three single-worker Vite templates now build it inside the component, per render tree, matching `templates/next`.

**A dev-only bug on the same lines.** Their SSR fallback origin pointed at `http://localhost:8787`, copied from the two-worker templates where a separate `wrangler dev` really does listen there. These compose Lunora into the SSR worker via `virtual:lunora/worker`, so nothing serves that port and every server-side query in dev hit a closed socket — invisible from the browser, which uses `location.origin`. A `vite.config.ts` plugin (ported from `examples/tanstack-start`) now defines `VITE_LUNORA_URL` from Vite's *resolved* port, so `--port` and a second app on the same machine both work.

**Missing declarations:** `vite/client` in three tsconfigs; an undeclared `h3` import in analog's Nitro route; a vestigial `/// <reference types="vitest" />` with no vitest; `@react-router/dev` in a `types` slot where it has no root entry point — that last one was aborting the semantic pass and masking two further errors.

**The gate then tightened.** Typechecking prefers a template's own `typecheck` script when it has one — a bare `tsc` skips `react-router typegen` and every `./+types/<route>` fails to resolve. With every typecheckable template at zero, the two-tier `lunora/auth-ui/**`-only logic is deleted: any diagnostic now fails the run.

## Verification

Full matrix under the strict gate: **9 PASS, 0 FAIL**.

The gate was proven to fail, not just to run — injecting one bad assignment into `lunora/messages.ts` and one into `src/server.ts` reports both and exits 1, which is what shows both trees are actually in scope.

## Not covered

`astro`, `nuxt` and `sveltekit` still have no in-scaffold typecheck (they need `@astrojs/check` / `vue-tsc` / `svelte-check`); the matrix prints that list on every run so a green result cannot be read as more than it is.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved template validation by running code generation and type checks for templates without build scripts.
  * Buildable templates now fail when any TypeScript diagnostics are detected during authentication UI checks.
  * Validation failures now identify the specific stage that failed.

* **Documentation**
  * Updated template validation documentation to describe the enhanced checks and failure reporting.

### DIFF
--- a/scripts/template-build-smoke.sh
+++ b/scripts/template-build-smoke.sh
@@ -193,11 +193,28 @@ run_typecheck() {
         } catch { process.stdout.write('no'); }
     " 2>/dev/null)"
 
+    # Returns the checker's exit status — callers need it. A `|| true` here would
+    # make "the checker refused to run" (missing binary, a failed `lunora codegen`
+    # inside a template's own `typecheck` script) indistinguishable from a clean
+    # pass, because neither produces an `error TS` line to grep for.
     if [[ "$has_script" == "yes" ]]; then
-        (cd "$scaffold_dir" && pnpm run typecheck 2>&1) > "$log" || true
+        (cd "$scaffold_dir" && pnpm run typecheck 2>&1) > "$log"
     else
-        (cd "$scaffold_dir" && pnpm exec tsc --noEmit 2>&1) > "$log" || true
+        (cd "$scaffold_dir" && pnpm exec tsc --noEmit 2>&1) > "$log"
     fi
+}
+
+# ---------------------------------------------------------------------------
+# A nonzero typecheck exit is only interesting when NOTHING was reported: `tsc`
+# exits 1 whenever it finds type errors, and those are gated (and printed) by the
+# caller. An exit with no diagnostic at all means the checker never got as far as
+# reading source, so a green result would prove nothing.
+# ---------------------------------------------------------------------------
+typecheck_never_ran() {
+    local status="$1"
+    local log="$2"
+
+    [[ "$status" -ne 0 ]] && ! grep -q 'error TS' "$log"
 }
 
 is_typecheck_unsupported() {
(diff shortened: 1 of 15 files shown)


prisis:
Fixed in 9757af9 — the finding was right: `run_typecheck` swallowed its status with `|| true`, so a checker that refused to start produced no `error TS` line and read as a clean pass.

I implemented it slightly differently from the proposed diff. Failing on *any* nonzero status would also fire on the normal case — `tsc` exits 1 whenever it finds type errors — and would replace the printed diagnostics with "typecheck command failed", losing the useful output.

So the status propagates, and a nonzero exit is only treated as a failure when the log carries **no** diagnostic at all:

```bash
typecheck_never_ran() {
    local status="$1"
    local log="$2"

    [[ "$status" -ne 0 ]] && ! grep -q 'error TS' "$log"
}
```

Real type errors keep falling through to the existing gate, which prints them. Verified against all three cases: exit 127 with no diagnostic → `FAIL(typecheck:never-ran)`; exit 1 with a diagnostic → falls through to the error gate; exit 0 → not flagged.

PR #406, opened 2026-08-11
fix(do): size subscription frames before sending them

prisis:
## Summary

Live-query fan-out (`@lunora/shard-engine`, `@lunora/do`) was sending more
bytes than it needed to, in two distinct ways. Both are fixed here, plus an
advisor lint (`@lunora/advisor`, `@lunora/codegen`) for the third — the one
only the app author can fix.

### 1. Delta-vs-snapshot was decided by row count

A refresh went out as one `{type:"delta"}` frame per changed row whenever
the diff was expressible, capped only by *"no more deltas than rows"*.

Every delta frame re-pays the whole `{"type":"delta","id":…}` + `"table":…`
+ cursor/epoch/watermark envelope around **one** row body — which the
snapshot pays once for the entire list. So the deltas stop being the cheaper
encoding well before they stop being a valid one, and a row count cannot see
where. Measured, 100-row list of ~290-byte rows:

| changed | snapshot | deltas | ratio |
| ------- | -------- | ------ | ----- |
| 1       | 29,060 B | 438 B  | 0.02x |
| 50      | 29,060 B | 21,980 B | 0.76x |
| 100     | 29,060 B | **43,970 B across 100 frames** | **1.51x** |

The cap was blind in the other direction too: a single delta into a short
list of small rows already costs more than re-sending the list, and it
allowed that as well.

Now both candidates are rendered and the smaller one is sent — stopping the
render as soon as the running delta total reaches the snapshot, so a
near-total change does not allocate envelopes that lose. `subscriptionFrames`
owns the frame layout and the choice together, so the sizing cannot drift
from the format — it *is* the format. Ties go to the snapshot, since one
frame is also one `ws.send` and one client apply.

`subscriptionListDeltas` keeps only its four expressibility conditions.
Whether sending the diff is worth it was never something a function with no
view of the wire could answer.

### 2. `unbounded_collect` (new advisor lint)

`filter_without_index` gates on `hasFilter`, so the widest read of all —
`ctx.db.query("t").collect()` with no index and no filter — fell straight
through it. The codegen feeder made that structural: it discarded every
discovered read without a `.filter()`, so no lint could have seen one.

An unindexed read records a whole-table dependency, so the refresh gate can
never skip it: every write to the table re-runs the query and re-sends the
whole table to every subscribed socket. Two of this repo's own examples trip
it, both on live-subscribed list queries.

The feeder now reports every read plus the materializing terminal, matched
against a known terminal set rather than the last chained method —
`.collect().then(...)` keeps the chain walk going, so the literal last call
is a combinator. `withGeoIndex` also joins the narrowing methods: it was
missing while only filtered reads were reported (a geo query is normally
written without one), and would otherwise have flagged the exact chain
`geo_index_unused` tells authors to write.

## Linked issues

None — found by a WebSocket payload audit of the subscription fan-out path.

## Test plan

- [x] `pnpm run lint:affected:types` — 73 projects, 0 errors
- [x] `pnpm run lint:eslint` — `@lunora/do`, `@lunora/shard-engine`, `@lunora/advisor`, `@lunora/codegen`
- [x] `pnpm exec prettier --write` on every touched file
- [x] `@lunora/do` 537 tests, `@lunora/shard-engine` 1123 (+8 new), `@lunora/codegen` 1202, `@lunora/advisor` 482
- [x] `pnpm run api:check` — 47/47 snapshots match (3 updated in this PR: `sendDeltaFrames` → `subscriptionFrames`, plus the new `terminal` field on `QueryReadIR` / `AdvisorQueryRead`)
- [x] Durable Object code changed → the workerd-pool `@lunora/do` suite is the one above, all green
- [x] Schema/codegen changed → every `examples/*` and `apps/playground` regenerated; only the two expected advisory diffs (`examples/chess`, `examples/feedback-board`), and nothing further after the rebase onto `alpha`
- [x] `pnpm --filter @lunora/shard-engine exec vitest bench` — the new bench runs clean

Reviewers should re-run `pnpm run api:check` and the `@lunora/do` suite.

## Checklist

- [x] Updated relevant docs in `apps/docs` (or `README.md` for package-local docs) — `packages/advisor/docs/index.mdx` gained the `unbounded_collect` row; the `apps/docs` copy is generated from it by `copy-package-docs.js` and is gitignored
- [x] If a new package was added: n/a
- [x] If migration impact: see the breaking-change note below

## Notes for reviewers

**Start with `packages/shard-engine/src/subscription-delivery.ts`** — the
whole delta-vs-snapshot decision lives in `subscriptionFrames`, and
`pushSubscriptionData` now just sends what it is handed.

**Breaking (pre-1.0, `alpha`):** `sendDeltaFrames` is removed from
`@lunora/shard-engine` in favour of `subscriptionFrames`, which returns the
frames rather than sending them. `subscriptionListDeltas` no longer returns
`undefined` for a near-total change — it decomposes any expressible diff.
Both are internal to the shard/engine boundary; no app code imports them.

**Frame-type flips are one-directional.** The new predicate is a strict
subset of the old one, so no socket that previously got a snapshot can now
get deltas — only the reverse. That is what keeps the seven non-JS SDKs
under `sdks/` safe: they handle `"data"` and `"delta"` in one arm and replace
wholesale, so this strictly reduces their exposure to that pre-existing
limitation.

### Deferred: paginated queries never diff

`.paginate()` returns `{ page, isDone, continueCursor }` — an object, so both
`subscriptionListDeltas` and the client's `applyDelta` reject it. A one-row
edit re-sends the entire page (measured 7,279 B for a ~290 B change, 25x),
and `usePaginatedQuery` holds **one subscription per loaded page**.

Not fixed here because it cannot be done safely without protocol work: a
client whose `applyDelta` bails replaces the whole query value with the raw
delta object, so page-deltas need a capability announced on the `connect`
frame and honoured by all seven SDKs. It is the largest remaining over-send
and worth doing, but as its own change. The new bench keeps a case measuring
what the rejection costs today.

### Measured, not changed

`subscriptionFrames` cost scales with **list length**, not with how much
changed: ~0.74 ms for a one-row edit to a 200-row list, paid per
`(socket, subscription)`. At 100 subscribers that is ~74 ms of
single-threaded shard time per write. Inherent to the current design
(identity-dependent queries cannot share a result under RLS); the bench pins
it so a regression shows up.

https://claude.ai/code/session_01AyG8dyPFieGYzRd7y1MMme

### DIFF
--- a/api-snapshots/advisor.api.md
+++ b/api-snapshots/advisor.api.md
@@ -520,6 +520,7 @@ interface AdvisorQueryRead {
     hasIndex: boolean;
     line: number;
     table: string;
+    terminal?: string;
 }
 ```
 
--- a/api-snapshots/codegen.api.md
+++ b/api-snapshots/codegen.api.md
@@ -558,6 +558,7 @@ interface QueryReadIR {
     hasIndex: boolean;
     line: number;
     table: string;
+    terminal?: string;
 }
 ```
 
--- a/api-snapshots/shard-engine.api.md
+++ b/api-snapshots/shard-engine.api.md
@@ -4152,12 +4152,6 @@ const selectShapeMemberIds: (sql: SqlExec, table: string, effectiveWhere: WhereI
 const selectShapeRows: (sql: SqlExec, table: string, effectiveWhere: WhereInput | undefined) => ShapeRow[];
 ```
 
-### `sendDeltaFrames` (const)
-
-```ts
-const sendDeltaFrames: (ws: FrameSink, subId: string, deltaFrames: ReadonlyArray<string>, cursorSuffix: string, lastMutationId?: number) => boolean;
-```
-
 ### `serializeSqlValue` (const)
 
 ```ts
@@ -4200,6 +4194,12 @@ const stableStringify: (value: unknown) => string;
 
 ```ts
 const stableWireKey: (value: unknown) => string;
+```
+
+### `subscriptionFrames` (const)
+
+```ts
+const subscriptionFrames: (input: SubscriptionFrameInput) => string[];
 ```
 
 ### `subscriptionListDeltas` (const)
--- a/examples/chess/lunora/_generated/shard.ts
+++ b/examples/chess/lunora/_generated/shard.ts
@@ -372,6 +372,26 @@ const LUNORA_TTL_SWEEPS: Array<{ after?: number; field: string; softDeleteField?
 /** Static schema advisories (computed by @lunora/advisor at codegen time) served via `__lunora_admin__:getAdvisories`. */
 const LUNORA_ADVISORIES: AdvisoryFinding[] = [
     {
+        "cacheKey": "unbounded_collect:players:32:profiles",
+        "categories": [
+            "PERFORMANCE"
+        ],
+        "description": "A query calls `.collect()` with no `.withIndex()` and no `.filter()`, so it materializes every row of the table. Any live subscription over it also re-sends that whole result to every subscribed client on every write to the table.",
+        "detail": "Query on \"profiles\" at players:32 calls .collect() with no index and no filter — it loads every row of \"profiles\" from the root Durable Object's SQLite into memory. A live subscription over this query records a whole-table dependency, so every write to \"profiles\" re-runs it and re-sends the full result to each subscribed socket.",
+        "facing": "EXTERNAL",
+        "level": "WARN",
+        "metadata": {
+            "exportName": "list",
+            "file": "players",
+            "line": 32,
+            "shardKind": "root",
+            "table": "profiles"
+        },
+        "name": "unbounded_collect",
+        "remediation": "Narrow the read with `.withIndex(\"name\", (q) => q.eq(...))`, cap it with `.take(n)`, or page it with `.paginate(args.paginationOpts)` so neither the scan nor the subscription payload grows with the table.",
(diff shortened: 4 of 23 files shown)


prisis:
Folding this into #407 rather than merging separately.

#407 was stacked on this branch, so it already contained every commit here; it is now rebased onto `alpha` and carries all five:

- `perf(do): pick delta frames over a snapshot by measured size`
- `feat(advisor): flag an unindexed, unfiltered collect`
- `perf(shard-engine): stop rendering delta frames once the snapshot wins`
- `perf(do): diff paginated live queries instead of re-sending the page`
- `fix(do): send a snapshot when a delta would misplace an inserted row`

Reviewing them as one unit is more honest anyway: the last commit fixes an insert-ordering desync that this PR's delta path introduced the exposure for, and the two are hard to judge apart.

The review feedback here was applied — the bench `snapshotJson` hoist (thanks @coderabbitai) and the early-abort render both landed in `perf(shard-engine): stop rendering delta frames once the snapshot wins`, which is in #407.

Note for anyone comparing CI: the CodSpeed red on this branch flagged `broadcastWhisper to 1024 members` and an in-batch re-projection — neither is touched by any commit here (no whisper/batch/relation lines in the diff), and CodSpeed's own summary reports "Different runtime environments detected… may affect the accuracy of the results". Cross-runner noise, not a regression.

Closing in favour of #407.

PR #412, opened 2026-08-14
feat(solid): support solid 2.0 alongside 1.x

prisis:
`@lunora/solid` now peers on `solid-js@^1.9.0 || ^2.0.0-rc.0` and serves
both majors from a single build. The public surface is unchanged for
existing 1.x consumers.

Solid 2.0 removed the primitives this adapter was written against — `on`,
`onMount`, the single-argument `createEffect`, and `Context.Provider` — and
moved JSX typing out of core into `@solidjs/web`. Rather than fork the
package, everything version-specific is isolated in `src/solid-compat.ts`:

- `trackedEffect` — `createEffect(on(src, fn))` on 1.x, the split-phase
  `createEffect(compute, apply)` on 2.0. Callers return their teardown.
- `onMounted` — `onMount` + `onCleanup` on 1.x, `onSettled` on 2.0.
- `providerOf` — probes for `.Provider`, since in 2.0 the context object
  *is* the provider component.
- `SolidChildren` / `SolidElement` — version-neutral element types, because
  `solid-js` exports the `JSX` namespace on 1.x only.

Version-specific members are read off the `solid-js` namespace object, never
imported by name: `import { on }` is an ESM link-time error under 2.x and
would break the package for every 2.x consumer regardless of code path.

The two JSX files are gone — `lunora-provider` and `create-auth` now build
their components with `createComponent`, which both majors export from the
package root. A `.tsx` source would have forced two builds, since the majors
compile JSX against different runtimes.

`tests/solid-v2-adapter` is the guard for the other half of that promise: it
installs Solid 2.0 as its only Solid and exercises the shipped bundle, and
asserts which major the adapter actually linked against so it can never pass
against the wrong one.

## Summary by CodeRabbit

* **New Features**
  * Added Solid 2.0 support across authentication, account settings, organizations, OAuth, two-factor authentication, and user controls.
  * Added Solid 2.0 project detection and dedicated CLI scaffolding templates.
  * Added compatibility with Solid 1.x and Solid 2.x applications.

* **Bug Fixes**
  * Prevented actions when required session or user details are unavailable.
  * Fixed organization member removal and improved avatar upload accessibility.
  * Updated verification error actions with clearer retry messaging.

* **Documentation**
  * Added Solid compatibility guidance, setup instructions, import differences, and quickstart examples.
  * Clarified template selection for Solid SPA projects.


prisis:
Review comments addressed in 186e045..68889ad, plus two defects a live run turned up.

**Fixed as asked**

- `features.ts` semver detection — `leadingMajor` replaced with semver's `minVersion`, guarded by `validRange`. `>1` now floors at 2; `2 || 1` and `^1.9.0 || ^2.0.0-rc.0` floor at 1. Policy for multi-major ranges is explicit: the manifest doesn't say which side the install resolved to, so the caller keeps its existing default and `@solidjs/web` stays the positive Solid 2 signal. Unparseable specifiers (`workspace:*`, `catalog:`, git URLs) decide nothing instead of throwing. Regression tests cover all six cases, and the mirrored ladder in `template-build-smoke.sh` tracks the same rule.
- `verifyEmailFailed` — reworded to the retry the card actually offers.
- Flow gates moved above `createController` in the Solid 2 cards (8 sites, incl. `BackupCodesCard`). This restores the order the provider's own doc comment specifies and `ConsentCard` already followed.

**Not applied**

- `untrack` around `ProfileCard`'s initial prop reads. `createController` invokes the factory from the component body, which Solid runs untracked (`createComponent` → `untrack`), so there is no tracking scope to escape and no dev warning to avoid — `untrack` would be a no-op. The snapshot-once semantics are already documented on `createController` itself ("a Solid component body runs once, so any options the factory closes over are captured at creation").

**Found by running both majors live** (scaffolded `solid-v2` and `tanstack-start-solid`, real dev server + workerd, driven in a browser)

- `createConnectionStatus` subscribed in the component body. `onConnectionStatus` fires its listener synchronously, so the write landed in an owned scope — Solid 2 raises `REACTIVE_WRITE_IN_OWNED_SCOPE`, and because the throw escapes render it halts the reactive system for the whole page. **Any Solid 2 app calling this rendered blank.** Now registered at mount.
  Both Solid fake clients returned without invoking the listener, unlike the real client and unlike the React mock — that divergence is why the suites were green against a crashing adapter. Fixing the fakes alone makes the existing Solid 2 test fail on the old code.
- The `solid-v2` template passed `allowUnauthenticatedShardAccess` to the Vite plugin, which only threads it into the generated `virtual:lunora/worker` composition. This template ships its own `src/server.ts`, so the flag reached nothing and **every sharded WebSocket was refused** — the template's own `.shardBy("channelId")` demo could not connect. Moved onto `defineApp().extend(...)`, matching the standalone template.

**Verified after the fixes** — both majors, live: socket `connected`, query loaded, mutation round-trips, subscription pushes the delta back (count 0 → 2, no refetch), zero console errors. The Solid 2 auth-ui payload renders too (sign-in/sign-up/2FA/backup-codes/profile/members), with gated-off cards correctly rendering nothing.

Green locally: `lint:types`, `api:check`, `lint:registry:sync`, `lint:package-json`, `dist:check`, prettier/eslint, the solid + auth-ui + cli + solid-v2 + auth-ui-solid-v2 + vis-templates suites, and `test:templates` for both Solid templates.

PR #425, opened 2026-08-18
feat(workflow): add declared workflow events

prisis:
## Problem

Cloudflare matches `instance.sendEvent({ type })` against `step.waitForEvent(name, { type })` by a bare string, and the payload crosses as `unknown`. Both ends drift silently:

- a typo'd type hibernates the instance until its timeout (24h by default) with no error anywhere;
- a payload whose shape changed resumes the workflow on garbage.

## What this adds

`defineWorkflowEvent(type, payloadValidator)` — one value both ends import, so there is no literal to typo and no second place to update on a rename.

```ts
// lunora/events.ts
export const orderApproved = defineWorkflowEvent("order-approved", v.object({ approvedBy: v.string() }));

// inside the workflow body — typed and parsed
const { approvedBy } = await ctx.waitForEvent(orderApproved, { name: "await approval", timeout: "7 days" });

// from a mutation/action — validated before the send
await ctx.workflows.get("orderPipeline").sendEvent(instanceId, orderApproved, { approvedBy });
```

- **Receive** (`ctx.waitForEvent`): resolves the wire type from the definition and parses the payload. A parse failure is non-retryable — the event is already consumed, so replaying the wait can only hibernate the instance again.
- **Send** (`WorkflowHandle.sendEvent(instanceId, event, payload)`): parses before the send, so a bad value fails the caller's request instead of waking the workflow. Addresses the instance by id, mirroring `ctx.agents.<name>.sendEvent(id, …)`.
- **Both boundaries re-check the definition** rather than trusting its brand (a public boolean), and reject the `lunora:` event-type namespace that the `ctx.parallel` branch join derives its own types from — so a definition-shaped object built from untrusted data cannot forge a branch outcome into a parent instance.

Additive: the native instance surface and the raw `ctx.step` API are unchanged, so the api-snapshot diff is additive too.

## Known sharp edge, documented

The default durable step name is `event:<type>`, which couples step identity to the wire type: renaming the event type also renames the step, and an instance that already recorded the wait replays into a fresh one nothing will satisfy. Docs carry a warn callout and every example passes a stable `{ name }`.

## Also here

A type-level lockstep test for `@lunora/server`'s hand-written mirror of the `ctx.workflows` surface. Nothing compared the two before, and codegen narrows `ctx.workflows` to `@lunora/workflow`'s own types as soon as an app declares a workflow — so the mirror is only exercised by apps with zero workflows and could rot unseen. It caught a stale mirror on its first run.

## Verification

`@lunora/workflow` 94 tests · server 570 · runtime 907 · codegen 1254 · platform-node 144 · `lint:affected:types` (73 projects) · `lint:affected:eslint` (58) · `lint:package-json` · prettier · api snapshots regenerated from a fresh build.

https://claude.ai/code/session_01B4ajaDyjxe3homh1sKVB5h

## Summary by CodeRabbit

- **New Features**
  - Added declared workflow events with typed payload validation.
  - Workflows can wait for events with optional names and timeouts.
  - Added validated event sending to workflow instances.
  - Added safeguards for invalid and reserved event types.
  - Exported event definitions and related helpers for integration.

- **Documentation**
  - Expanded guidance for declaring, waiting for, and sending events.
  - Documented validation, failure behavior, stable wait names, and instance management APIs.
  - Updated package metadata to describe Worker-scoped Access policies and verified Access tokens.

### DIFF
--- a/api-snapshots/lunora.api.md
+++ b/api-snapshots/lunora.api.md
@@ -905,6 +905,10 @@ Re-exported from `@lunora/server` — signature tracked at its source.
 
 Re-exported from `@lunora/server` — signature tracked at its source.
 
+### `WorkflowEventDefinition` (interface)
+
+Re-exported from `@lunora/server` — signature tracked at its source.
+
 ### `WorkflowHandle` (interface)
 
 Re-exported from `@lunora/server` — signature tracked at its source.
@@ -5415,6 +5419,10 @@ Re-exported from `@lunora/server` — signature tracked at its source.
 
 Re-exported from `@lunora/server` — signature tracked at its source.
 
+### `WorkflowEventDefinition` (interface)
+
+Re-exported from `@lunora/server` — signature tracked at its source.
+
 ### `WorkflowHandle` (interface)
 
 Re-exported from `@lunora/server` — signature tracked at its source.
@@ -7078,6 +7086,10 @@ Re-exported from `@lunora/server` — signature tracked at its source.
 Re-exported from `@lunora/server` — signature tracked at its source.
 
 ### `WorkflowCreateOptions` (interface)
+
+Re-exported from `@lunora/server` — signature tracked at its source.
+
+### `WorkflowEventDefinition` (interface)
 
 Re-exported from `@lunora/server` — signature tracked at its source.
 
--- a/api-snapshots/server.api.md
+++ b/api-snapshots/server.api.md
@@ -2440,6 +2440,16 @@ interface WorkflowCreateOptions<Params = Record<string, unknown>> {
 }
 ```
 
+### `WorkflowEventDefinition` (interface)
+
+```ts
+interface WorkflowEventDefinition<Payload = unknown> {
+    readonly isLunoraWorkflowEvent: true;
+    readonly payload: Validator<Payload>;
+    readonly type: string;
+}
+```
+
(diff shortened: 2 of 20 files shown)


prisis:
All three review findings fixed in b0779a6dc:

- **`isWorkflowEventDefinition` brand-only check** — valid. The predicate hands the caller a `type` and a `payload.parse`, so it now runs the same `eventDefinitionProblem` check both boundaries use; a brand-only object no longer passes. Test added for that case.
- **`@lunora/server` devDependency** — valid. Pinned to `1.0.0-alpha.75`, matching codegen/seed/replica/cloudflare-access; `workspace:*` in this repo is for the unpublished internals only (`@lunora/dispatch`, `@lunora/search-core`).
- **Reserved-type docs wording** — valid, it overstated the guarantee. Now says `defineWorkflowEvent` and the typed `sendEvent` reject `lunora:` types, and that the native untyped `instance.sendEvent` is not covered.

Separately, the red **Lint (generated files)** check was not from this branch: `apps/docs/src/data/packages.ts` on `alpha` carries a cloudflare-access description that its curated source (`packages-metadata.json`) never received, so the generator cannot reproduce the committed file — it fails on any PR based on current alpha. The curated entry now matches and the generated output is byte-identical.

PR #426, opened 2026-08-19
feat(studio): close nine gaps in the data browser, SQL console, and dashboards

prisis:
Nine gaps in the Studio's data browser, SQL console and dashboards, plus a tenth split out along the way. Eight of the ten are wiring rather than building: the capability already existed somewhere in the repo and the consumer did not reach for it.

Plan: [`plans/363-studio-data-surface-parity.md`](plans/363-studio-data-surface-parity.md). The one genuinely new mechanism (a conversational assistant) is **not** in this PR — it produced its own plan instead, [`plans/364`](plans/364-studio-conversational-assistant.md).

## The headline defect

A dashboard widget could only ever be a bar chart. `dashboards-panel.tsx` rendered `<SqlResultChart result={result} />` with no `axes`, and the component reads `axes !== undefined && suggestedValue !== undefined ? axes.kind : "bar"` — so it took the constant arm on every render. Its line and area paths, and the assistant's chart inference, were both unreachable from the one surface whose purpose is charting.

## What changed

| | |
| --- | --- |
| **Dashboards** | Chart type is pickable and honoured. Four tile kinds — chart, single value, table, note — with drag reordering. |
| **Row editor** | Field widgets come from the schema instead of the runtime value: a string-literal union is a dropdown, a null boolean is a checkbox. |
| **Grid** | Paste a TSV block as staged edits. `v.storage()` cells preview the object. Headers carry a declared-type glyph. |
| **SQL console** | Multi-statement scripts run as N separately-gated calls, with a per-statement result strip. |
| **History** | Scoped to the tab by default, opt-in to disk. |
| **Settings** | Real preferences: both keyboard shortcuts are rebindable. |

## Two premises worth reading before reviewing

**Multi-statement rejection is deliberate, not an oversight.** `classifyStatement` refuses any `;` that is not a single trailing one, and that rule is the console's enforcement point — raw writes bypass the schema-aware writer and desync the doc-store's FTS, aggregate and rank shadow tables. So the split happens **above** the gate: each statement is classified locally and submitted as its own call, a `;`-joined string never reaches the server, and a refused statement is reported without being sent and does not abort the ones after it.

**The enum dropdown needed a codegen change.** `ColumnMeta` could not feed a dropdown because `buildTableColumns` wrote `type: resolved.kind` and discarded the union's members. It now emits `enumValues` when every member is a string literal, parsing `literalValue` back from the canonical source text — passing it through raw would offer every option with its quotes still on. All-or-nothing: a mixed union has legal values outside any list, and a numeric union would stage `"1"` where the column holds `1`. Both fall back to the old behaviour, as do `v.any()` and plugin-owned tables.

`ColumnMeta` moves in every place that declares it — codegen's emitted inline type, `@lunora/shard-engine`, the studio — in one commit. The field is optional, so a Studio talking to an older worker falls back rather than blanking.

## Two pre-existing defects found on the way, and fixed

Both predate this branch. Each got its own commit and its own tests, after the workstream that found it had landed — the plan's STOP condition forbids touching the gate in passing, so neither was fixed mid-workstream.

**The read-only gate read a quoted semicolon as a statement boundary.** The batch check was `indexOf(";")` over raw text, so **four** classes of legal read-only query were refused outright, not one: a `;` inside a string literal, a `--` line comment, a `/* */` block comment, or a `"…"` / `` `…` `` / `[…]` quoted identifier. SQLite's parser never sees a second statement in any of those.

The scan now runs on a same-length masked copy, so a rejection's `offset` still indexes the caller's own string. Anything unclosed returns `undefined` and falls back to the raw scan that shipped — such a statement is a syntax error to SQLite anyway, and masking to end of input would hide a real `;` behind a stray quote. Doubled delimiters are handled as the escapes they are.

**Deliberately not extended to the forbidden-keyword scan.** Rejecting a mutating word even inside a literal is documented on that constant as an accepted trade for a tool that must never corrupt the doc-store's shadow tables. Masking the batch scan fixes a rule nobody chose; that one someone did. A test pins it.

**The studio's "possible request loop" warning was the detector's bug.** A probe counting admin reads across a single mount found `describeTables`, `listTables`, `aiAvailable` and `maskPolicies` each issued exactly **once** — there was no loop. The detector's bucket is module-level and never reset, so it cannot tell one component re-issuing 25 times from 25 components mounting, and its comment justified the threshold with "TanStack dedupes identical in-flight reads" — true only of reads that overlap, never of sequential mounts each awaited to completion, which is exactly what a suite produces. The test environment is now excluded and the claim corrected.

## Four places the plan was wrong

Recorded inline in `plans/363` as each workstream landed:

- `useRunSql` is the **dashboard's** hook, not the console's — returning a list there would have forced `results[0]` on every tile, the exact special-casing the plan wanted to avoid. `script` sits alongside `result` instead.
- Flipping the history-persistence toggle **on** would have discarded the history it was asked to keep, because `usePersistedList` reloads from the new storage area. It now seeds the destination before flipping.
- A text dashboard tile needs its own component: a conditional hook is not available, and a text tile that still issues `runSql` is a query nobody asked for on every render.
- There are four declarations of the `ColumnMeta` shape, not three.

## Verification

`@lunora/studio` 1063 tests · `@lunora/codegen` 190 tests · `lint:types` + ESLint clean · `api:check` green across 48 snapshots after `api:update` (one line: `ColumnMeta.enumValues`) · `lint:package-json` green · golden codegen fixtures and all 13 example `_generated` regenerated in their own commit.

Two of the examples already declare `v.union(v.literal(…))` columns, so the regeneration exercised the emitter end to end on real schemas rather than only the hand-built IR in the unit tests — `chess.games.status` emits `["active", "completed", "abandoned"]`.

https://claude.ai/code/session_01WYuhTqTiAJsw5NdF5i1QSQ

## Summary by CodeRabbit

* **New Features**
  * Added multi-statement SQL execution with per-statement results, errors, and selection.
  * Added configurable SQL history persistence for the session or browser.
  * Expanded dashboards with chart, KPI, table, and text tiles, including drag-and-drop reordering.
  * Added TSV paste support, schema-aware editors, enum dropdowns, type indicators, and bucket-aware storage previews.
  * Added customizable keyboard shortcuts and chart type selection.
* **Bug Fixes**
  * Improved validation of pasted values and handling of failed dashboard queries.
  * Improved nullable enum handling and preservation of unknown values.
  * Preserved text-editing shortcuts while using configurable commands.
  * Improved SQL diagnostics and read-only validation for multi-statement scripts.


packages/studio/src/features/data/data-browser-grid.tsx
```diff
@@ -697,6 +865,7 @@ const RowSelectCell = ({ row }: { row: Row<TableRow> }): ReactElement => {
  * callbacks so this stays a pure render of the page's rows.
  */
 const DataBrowserTableView = ({
```
prisis, line 867: Leaving this open deliberately rather than resolving it.

`DataBrowserTableView` is 313 lines — 13 over the threshold. The natural seam is `renderRow`, which closes over **nine** values (`tableRows`, `columnSlice`, `pinOffsets`, `edit`, `editable`, `active`, `setActive`, `rowId`, `rowKey`). Extracting it means threading nine props through the hottest path in the file — it runs per virtualised row — in code that landed in this PR.

That trade reads badly right now: real regression risk in the most heavily exercised part of the change, to satisfy a line count, and a mechanical split wouldn't make the component easier to follow. The rule is right; this isn't the moment.

Better as its own change, where the split can be designed (a `GridRow` with a narrow props contract, or a context for the row-scoped values) and reviewed on its own merits.

PR #436, opened 2026-08-21
fix(x402): parse release amounts, slim deps

prisis:
- `releaseSpendOnFailure` now parses the server-controlled requirement amount with the same fail-closed `parseAtomicAmount` the guard and selection filter use. Previously a junk amount made the failure hook throw (masking the original payment-creation error) and a negative string inflated the spend ledger via `release()`, locking the wallet out for the rest of the run. An unparsable amount now releases nothing.
- Restored coverage thresholds in the package's vitest config. The inline mirror of the shared coverage block omitted `thresholds`, so this was the one package where coverage had no gate; it now spreads `DEFAULT_COVERAGE_THRESHOLDS` (branches 70, functions 80, lines 80, statements 80). Measured coverage is ~95% on all four metrics, so no ratchet was needed.
- Moved `@x402/evm`, `@x402/svm`, `viem`, and `@solana/kit` from hard dependencies to optional peers (devDependencies for the package's own build/tests), matching the existing `@coinbase/cdp-sdk` treatment. The scheme modules were already lazily `import()`-ed per network family, but the manifest still forced every install to download both chain toolchains.
- Added a single `importOptionalPeer(load, guidance)` helper that all seven lazy peer imports (including the pre-existing CDP one) route through. It converts a *missing* module into an `ENV_INVALID` naming the peer and the alternative, but rethrows any other failure untouched — an installed-but-broken peer (runtime incompatibility, bad transitive dep, a throw at evaluation time) must report its own cause, not "install it". Detection matches the error code or message and walks the `cause` chain, since loaders and test runners wrap the resolution failure; the original error is attached as `cause` either way.
- `resolveEvmAccount` now returns the structural `ClientEvmSigner` instead of viem's `PrivateKeyAccount`, which removes viem from the published declarations so a consumer without the peer still type-checks. The pay rail's `ClientSvmSigner` still resolves through `@solana/kit`'s transitive `@solana/accounts` (pre-existing: packem inlines `@x402/svm`'s alias regardless of how the dependency is declared), so the README now states which peers each network family needs at both install and compile time.

**Breaking (pre-1.0 alpha):** consumers must install the peers for their chain family — `@x402/evm` + `viem` for EVM, `@x402/svm` + `@solana/kit` for Solana. Missing a peer fails with an `ENV_INVALID` error naming exactly what to install. No in-repo examples or templates depend on `@lunora/x402`, so none needed updating.

## Verification

- `pnpm --filter "@lunora/x402" run test` — 9 files, 112 tests, all pass (new cases: negative/junk release amounts, missing EVM and SVM peers on both rails, and a broken-but-installed peer surfacing its own error)
- `pnpm --filter "@lunora/x402" run test:coverage` — passes at the restored thresholds; a deliberately-too-high threshold was confirmed to fail the run, then reverted (gate is live)
- `pnpm --filter "@lunora/x402" run lint:types` / `lint:eslint` — clean
- `pnpm run lint:package-json` — all 76 manifests sorted
- `pnpm run build:packages` + `pnpm run api:check` — public API matches all 48 snapshots (note: `@lunora/x402` has no committed snapshot of its own)
- `pnpm install` — lockfile regenerated; `pnpm peers check` reports no x402-related issues
- Published declarations inspected after a rebuild: no `viem` import remains

## Summary by CodeRabbit

* **New Features**
  * Web3 integrations are now optional, allowing installations to include only the protocol, EVM, Solana, or custody features they need.
  * Clear setup guidance is provided when an optional integration is missing.

* **Bug Fixes**
  * Invalid or negative payment amounts no longer modify spend tracking or cause unexpected failures.
  * Integration-loading errors now provide more actionable configuration guidance.

* **Documentation**
  * Added installation instructions, TypeScript setup notes, missing-dependency guidance, and bundled-license attributions.


prisis:
@coderabbitai review

prisis:
@coderabbitai review

PR #441, opened 2026-08-21
fix(server): harden validation, presence, filters

prisis:
- **Output-schema mismatches are internal errors on the RPC path.** A builder procedure with `.output()` let the raw `ValidationError` (a 400 whose message embeds the offending server-side value) escape to the client. The REST path's `applyOutput` helper — which re-tags the mismatch as `INTERNAL_SERVER_ERROR`, redacted at the wire and logged server-side — now lives in `src/apply-output.ts` and both transports route through it. Args validation failures stay 400s.
- **`contains` is offered only on string-typed filter columns.** `defineListArgs` hardcoded `contains` into every column's operator object, so a published numeric/boolean/timestamp column accepted `?where[age][contains]=4` — semantically void and a non-sargable scan. The operator object now declares it only for string-typed columns, and the exported `toQueryArgs` sanitizer drops it elsewhere. String-typed includes enum columns (`v.union` of string literals) and bare `v.literal`s — judging those by `kind` alone would strip the operator and, since an emptied predicate is dropped, silently return the *unfiltered* set. **Breaking:** previously-accepted `contains` on non-string columns is stripped (consistent with the module's allow-list mechanism) and never reaches the SQL compiler.
- **One secret-key rule, in `shared/`.** Four copies of the "does this key name imply a secret" regex existed — the runtime redactor, the `.dev.vars` scaffolder, `lunora deploy` and `lunora doctor` — kept in step only by a comment, and they had drifted. They are now one zero-dep, bundler-inlined definition in `shared/secret-key.ts`: a case-insensitive suffix match, so the spellings that actually appear in request bodies and thrown errors (`password`, `apiToken`, `authSecret`) redact, as do no-separator compounds (`OPENAI_APIKEY`, `APITOKEN`) and Title-case keys. Over-redaction is the documented, deliberate failure direction — `MONKEY` and `APIKEY` are structurally identical, so masking a non-secret (one confusing log line) is chosen over missing a credential.
- **`storageRules`' one synchronous guarded method is documented and pinned.** `getUrl` is the only non-Promise member of the guarded surface; a future async refactor of the wrapping loop would silently turn `ctx.storage.getUrl` into a Promise for guarded procedures only. A comment states the invariant and a test asserts the wrapped call returns a plain string. No behaviour change.
- **`listPresent` is bounded and presence self-reaps.** The query collected every row a room ever accumulated (the TTL filter hides stale rows but never deletes them, and nothing schedules `sweep` by default). A new `(roomId, lastSeen)` index plus a `maxSessions` option makes the read newest-first with a hard cap, and each heartbeat reaps up to 8 of its room's oldest rows using a cutoff a full `max(grace, ttl)` window behind the visibility cutoff — a row the read filter could still show, or a grace-window reconnect could revive, is never deleted. The cap counts session rows (one per open tab) and the per-user dedup runs after the read, so it is named and documented for sessions and defaults to 1024; a non-finite value falls back to the default rather than reaching the reader as `LIMIT NaN`. `sweep` remains as optional bulk hardening.

## Verification

- `pnpm --filter "@lunora/server" run test` — 590 passed (24 new tests across builder, list-args, env, storage-rules, presence)
- `pnpm --filter "@lunora/config" run test` — 596 passed; `pnpm --filter "@lunora/cli" run test` — 1257 passed (both consume the shared secret-key rule)
- `lint:types` + `lint:eslint` clean for `@lunora/server`, `@lunora/config`, `@lunora/cli`; `pnpm run lint:package-json` clean
- `pnpm run build:packages && pnpm run api:check` — all 48 snapshots match (`DefinePresenceOptions.maxSessions` recorded)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistent detection and redaction of secret-like environment keys across deployment, diagnostics, configuration, and server workflows.
  * Fixed filtering so `contains` is available only for supported string fields, while invalid predicates are safely excluded.
  * Improved handling of invalid server output, which now surfaces as a clear internal error.
* **Improvements**
  * Presence heartbeats now clean up stale sessions and support configurable session limits.
  * Added safer, bounded session reads for presence listings.

### DIFF
--- a/api-snapshots/server.api.md
+++ b/api-snapshots/server.api.md
@@ -304,7 +304,7 @@ interface DefinePolicyInput<Context = unknown> {
 ```ts
 interface DefinePresenceOptions {
     disconnectGraceMs?: number;
-    maxMembers?: number;
+    maxSessions?: number;
     ttlMs?: number;
 }
 ```
--- a/api-snapshots/server.api.md
+++ b/api-snapshots/server.api.md
@@ -304,6 +304,7 @@ interface DefinePolicyInput<Context = unknown> {
 ```ts
 interface DefinePresenceOptions {
     disconnectGraceMs?: number;
+    maxMembers?: number;
     ttlMs?: number;
 }
 ```
--- a/packages/cli/src/commands/deploy/handler.ts
+++ b/packages/cli/src/commands/deploy/handler.ts
@@ -28,6 +28,7 @@ import { join } from "@visulima/path";
 import { Spinner } from "@visulima/spinner";
 import { Project } from "ts-morph";
 
+import { isSecretKeyName } from "../../../../../shared/secret-key";
 import { evaluateAdvisoryGate, resolveStrictAdvisories } from "../../util/advisory-gate";
 import type { ApiSpec } from "../../util/api-spec";
 import { parseApiSpec } from "../../util/api-spec";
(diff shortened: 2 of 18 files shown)


prisis:
@coderabbitai review

PR #442, opened 2026-08-21
fix(auth): close silent no-ops in admin paths

prisis:
Closes silent no-op and fail-closed paths in `@lunora/auth`'s admin plane and audit trail, fixes DO-mode route capture, and untangles the `@lunora/server` dependency.

- **`setUserPassword` no longer silently no-ops.** It 404s an unknown user (`USER_NOT_FOUND`) and creates the credential account when the user has none (OAuth-only, or created without a password) — previously the underlying `updatePassword` matched zero rows and reported success while the user still could not sign in. It checks existence before paying for the KDF, and refuses (`EMAIL_PASSWORD_DISABLED`) rather than planting a credential account on a deployment with email+password turned off; rotating an existing credential stays allowed.
- **`withAuthAudit` forwards the caller's `hooks.after` return value.** better-auth treats any non-undefined after-hook return as the endpoint's new response; the composition discarded the caller's value, so an app hook that rewrote a response silently lost the rewrite once wrapped. The audit record now runs as a pure side effect.
- **DO-mode auth routing matches the base path at a segment boundary.** With base `/api/auth`, app routes like `/api/authorize` were forwarded into the auth Durable Object and answered 404 there. Both auth modes now share one module-internal predicate, `isAuthRoutePath` — no public API surface change.
- **`removeUser` cascades over user-keyed plugin tables.** `member`, `teamMember`, `passkey`, `twoFactor`, `oauthAccessToken`, `oauthRefreshToken`, `oauthConsent`, `deviceCode`, `walletAddress` and sent invitations are unwound (gated on `getAuthTables` presence, scoped to the removed user). None of these declare `onDelete` in better-auth's schema, so nothing removed them even on an FK-enforcing backend, and D1 does not enforce FKs at all — an operator deleting a user to revoke access left a live third-party bearer token authenticating as that userId, plus a TOTP secret and backup codes for an account that no longer exists.
  - Deliberately **not** cascaded: `ssoProvider` and `oauthClient` (their `userId` is the admin who registered a tenant-wide resource — removing them would break login for every other user of that domain or app), audit rows (forensics), and invitations addressed **to** the user (email-keyed).
  - `removeUser` now refuses to delete the **last owner of an organization** (`LAST_ORGANIZATION_OWNER`) instead of leaving an org alive that nobody can administer, invite to, or delete.
  - The cascade is several non-atomic writes ending in `deleteUser`, so a mid-failure leaves a partially-cascaded user. Every step is idempotent: re-running `removeUser` is the recovery.
- **A malformed audit `detail` cell degrades to one incomplete entry.** `readAuthAuditLog` no longer throws the whole page away when one row's `detail` is not valid JSON (or not a plain object); that entry reads back without its detail payload.
- **The audit table DDL is single-flighted per executor.** The CREATE plus a guaranteed-throwing ALTER ran before every append/read — on the request path of every audited auth request. A WeakMap memo (rejection-evicting, same shape as `migrate.ts`) runs them once per long-lived executor; a fresh executor still self-provisions.
- **`@lunora/server` is now a required peer range (`>=1.0.0-alpha.24 <2.0.0-0`) + dev pin instead of an exact-pinned dependency.** The exact pin gave consumers on a newer `@lunora/server` a second installed copy, splitting the `Middleware`/`TableDefinition` type identities in one build. Breaking: consumers must provide `@lunora/server` themselves (every Lunora app already does). The range matches the three sibling packages that already peer on `@lunora/server` (`cloudflare-access`, `replica`, `seed`).

## Verification

- `pnpm --filter "@lunora/auth" run test` — 32 files, 392 tests, all green (15 new)
- `pnpm --filter "@lunora/auth" run lint:types` and `lint:eslint` — clean
- `pnpm run lint:package-json` — clean
- `pnpm run api:check` — all 48 snapshots match; the auth snapshot is unchanged from `alpha`
- `node scripts/check-sibling-peer-ranges.js` and `check-peer-catalog-refs.js` — clean

### DIFF
--- a/api-snapshots/auth.api.md
+++ b/api-snapshots/auth.api.md
@@ -923,6 +923,12 @@ const eventForPath: (path: string) => AuthAuditEvent | undefined;
 const handleAuthRequest: (auth: LunoraAuth, request: Request, basePath?: string) => Promise<Response | undefined>;
 ```
 
+### `isAuthRoutePath` (const)
+
+```ts
+const isAuthRoutePath: (pathname: string, basePath: string) => boolean;
+```
+
 ### `loadEmailDomainLists` (const)
 
 ```ts
--- a/api-snapshots/auth.api.md
+++ b/api-snapshots/auth.api.md
@@ -923,12 +923,6 @@ const eventForPath: (path: string) => AuthAuditEvent | undefined;
 const handleAuthRequest: (auth: LunoraAuth, request: Request, basePath?: string) => Promise<Response | undefined>;
 ```
 
-### `isAuthRoutePath` (const)
-
-```ts
-const isAuthRoutePath: (pathname: string, basePath: string) => boolean;
-```
-
 ### `loadEmailDomainLists` (const)
 
 ```ts
--- a/api-snapshots/config.api.md
+++ b/api-snapshots/config.api.md
@@ -1327,6 +1327,12 @@ const readProjectRemotePreference: (projectRoot: string) => RemotePreference;
 const readProjectTarget: (projectRoot: string) => string | undefined;
 ```
 
+### `removeDevVariableLine` (const)
+
+```ts
+const removeDevVariableLine: (content: string, key: string) => string;
+```
+
 ### `requiredSecrets` (const)
 
 ```ts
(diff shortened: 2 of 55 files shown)


prisis:
@coderabbitai review

prisis:
@coderabbitai /review

PR #443, opened 2026-08-21
fix(cli): guard sdk vendoring and imports

prisis:
Four hardening fixes across the CLI (plus the shared studio-host layer in `@lunora/config` and its Vite consumer):

- **Refuse symlinks when vendoring SDK transports.** `lunora sdk generate` copied the transport directory with a `cpSync` filter that only skipped test files; symlinks were copied verbatim, so a hostile or compromised transport source could plant a link (e.g. `config.py -> ~/.ssh/id_rsa`) inside the user's project. The copy filter now `lstat`s every entry and throws naming the offending relative path, matching the registry copy-in guard's refuse-don't-skip stance. Regression test included.
- **First test suite for `lunora sdk generate`.** The command writes into the user's project and had zero tests. Five end-to-end cases driven through the offline `--from` seam (no network): happy path (vendored files with test files filtered, stamp contents, generated surface, untyped-result count), empty-methods warning, invalid spec naming its path, unknown `--lang` naming the supported list, and the transport-first ordering (a missing transport fails before any generated file or stamp lands).
- **Line-numbered validation for every import envelope.** `lunora import` validated NDJSON envelopes with a line-numbered error only when a storage/document rewrite was configured; the plain invocation forwarded raw lines, so a corrupted file failed as a whole-batch server error with no way to find the offending line in a multi-GB file. Every envelope is now parsed through the same validation; with no rewrite the original string still passes through untouched (no re-serialisation, no double parse).
- **Cache headers on the CLI studio dev server.** The non-Vite studio host served the unhashed studio entry/stylesheet/chunks with only a `Content-Type`, so browsers heuristically cached them and served a rebuilt `@lunora/studio` stale until a hard reload — and the token-embedding SPA document was heuristically cacheable, putting the admin token in the browser's disk cache. The revalidation decision (stamp-keyed weak ETag + `If-None-Match`) and the two cache policies now live in `@lunora/config/studio-host` (`asset-cache.ts`), following the transport-guard consolidation; the Vite host routes its behaviour-identical ETag/304 block through the helper, and the CLI host now sends `no-cache` + ETag with 304s on assets and `no-store` (never an ETag) on the token-bearing document. `api-snapshots/config.api.md` updated for the new exports.

## Verification

- `pnpm --filter "@lunora/cli" run test` — 95 files, 1269 tests, all pass
- `pnpm --filter "@lunora/config" run test` — 38 files, 601 tests, all pass
- `pnpm --filter "@lunora/vite" run test` — 18 files, 197 tests, all pass (studio-plugin tests unchanged)
- `lint:types` + `lint:eslint` green for all three packages
- `pnpm run build:packages && pnpm run api:check` — all 48 snapshots match

---

**Merge note — interacts with the other studio-host PR.** `packages/vite/src/studio-plugin.ts` imports `headerValue` from `@lunora/config/studio-host`. This PR removes some of its call sites and #443/#461 (whichever is the other one) removes the rest, so once **both** have landed the import is orphaned and `lint:types` fails with TS6133. Git will not flag it — there is no textual conflict. Whichever of the two lands **second** must drop `headerValue,` from that import block at rebase time.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and error reporting for malformed data-import envelopes.
  * Preserved envelope data more reliably during imports, including missing or unusual document values.
  * Added protection against symlinks in vendored SDK sources before copying files.

* **Performance**
  * Improved Studio asset caching with conditional `304 Not Modified` responses.
  * Prevented caching of the Studio HTML document to ensure fresh application loads.


prisis:
@coderabbitai /review

PR #444, opened 2026-08-21
fix(replica): correct pk affinity and id binds

prisis:
Three LocalMirror/diff-applier correctness fixes in `@lunora/replica`:

- **Normalize row-id binds in the diff applier.** The row id bound into every DELETE/UPDATE bypassed `normalizeBindValue`, so a non-bindable `change.id` (boolean/object/undefined in an untrusted server payload) threw inside `database.transaction()` and rolled back the entire batch, unrelated rows included. Both id binds now route through the normalizer like every other bound value.
- **Drop change subscribers on mirror close.** `close()` left `#changeListeners` attached, so anything that closes and rebuilds a mirror (schema reset, sign-out, HMR) kept the old subscribers — and the component trees their closures capture — alive on the dead mirror object, and a post-close event would notify components about a database that no longer exists. `close()` now clears the set.
- **Infer the primary-key column's affinity.** Affinity inference covered every non-PK column but left the pk hard-coded `TEXT`, so a numeric id sorted and range-filtered lexicographically (`ORDER BY id` put 10 before 9; `WHERE id > 5` compared strings). The create path now infers the pk affinity from the first observed id, falling back to TEXT.
  - A numeric pk is declared `INT`, **not** `INTEGER`: a column typed exactly `INTEGER PRIMARY KEY` is SQLite's rowid alias and accepts integers only, so a later non-numeric id for the same table would raise `datatype mismatch` inside the diff transaction and roll back the whole diff. `INT` carries the same INTEGER affinity without the alias, so heterogeneous ids still apply.

## Migration — persisted data

`MIRROR_SCHEMA_VERSION` bumps 2 → 3. `#reconcileSchemaVersion` drops **every mirrored table** on the first open of an existing mirror and re-seeds from subsequent diffs — the same mechanism v2 used, re-triggered here. Impact by consumer:

- `subscribeToMirror` consumers are unaffected in practice: each frame carries a full snapshot, so the mirror repopulates on the next update.
- **A consumer resuming from its own persisted cursor loses every existing row** and only catches up from that cursor forward. Such consumers should reset their cursor (or take a full snapshot) once on upgrade.

Release notes should carry this as a one-time re-seed.

## Verification

- `pnpm --filter "@lunora/replica" run test` — 362/362 pass; new real-engine cases in `adapters.test.ts` run across sql.js, better-sqlite3, and the sqlite-wasm double: non-bindable-id batch survival, post-close unsubscribe safety, numeric-pk `ORDER BY`/`WHERE`, string-pk still TEXT, delete-only and update-without-pk affinity inference, and a heterogeneous-id batch
- Negative checks: the non-bindable-id test fails against the pre-fix diff applier (better-sqlite3 throws), and the heterogeneous-id test fails on all three engines with `datatype mismatch` if the pk is declared `INTEGER` instead of `INT`
- `pnpm --filter "@lunora/replica" run lint:types` / `lint:eslint` — clean; branch coverage 89.09% (gate 88%)
- `pnpm run api:check` — all 48 snapshots match after a full `build:packages`


prisis:
@coderabbitai Fixed in 87aa869dc — `#inferPkAffinity` now narrows via the `change.type` discriminant like the sibling helpers; id-selection order and the TEXT fallback are unchanged.

PR #450, opened 2026-08-21
fix(notify): route push targets by their kind

prisis:
## What was wrong

The composite push provider routed an entire send by the kind of its **first** target: `pick(endpoints[0]).send(payload)`. The SSRF re-check loop directly above already iterated every target, and `notify.send()` hands a caller-shaped message straight to the engine — so a multi-recipient push with mixed kinds, `ctx.notify.send({ push: { to: [webPushSubscription, fcmToken] } })`, handed the FCM token to the Web Push provider (or the reverse): silent misdelivery for any mixed-kind fan-out. The single-target `ctx.push.send`/`broadcast` paths always pass one target, which is why no test caught it.

## What it does now

`routingPushProvider.send` partitions the targets by kind after the SSRF re-check:

- A single-kind `to` (every `ctx.push` call, and most `notify.send()` payloads) passes through to its provider unchanged — same behavior as before, including the directed unconfigured-channel errors.
- A mixed-kind `to` narrows the payload per group (preserving the scalar-vs-array shape of `to`), sends each group through its own provider, and merges the results: successful only when both sends are, returning the first failure otherwise — a failed group is never masked by a succeeding one.

## Verification

- `pnpm --filter "@lunora/notify" run test` — 120 passed (117 existing + 3 new: mixed `to` routes each target to its own provider stub, single-kind arrays pass through unchanged, a failing group is not masked by a succeeding one)
- `pnpm --filter "@lunora/notify" run lint:types` — clean
- `pnpm --filter "@lunora/notify" run lint:eslint` — clean
- `pnpm run build:packages && pnpm run api:check` — all 48 snapshots match (no public surface change)


prisis:
All three items confirmed against the code and fixed in `7dfd023`.

**1 — Partial delivery reported as total failure: CONFIRMED, FIXED.** `pick(undefined)` threw between the two sends, so an unconfigured `fcm` channel rejected the whole send *after* the web-push half was already POSTed. Both channels are now resolved before either send is issued, so an unconfigured channel fails with nothing delivered.

**2 — No isolation: CONFIRMED, FIXED — and it was worse than described.** The sends now run under `Promise.allSettled`. Note the `async` wrapper around each: `Provider.send` returns `MaybePromise`, so a provider that throws *synchronously* (the repo's own `mockThrowingPushProvider` does exactly this) escapes past `allSettled`, which only catches rejections. Without the wrapper the sibling group is never even called — my first pass at this fix still failed the isolation test for that reason.

**3 — Merge loses data: CONFIRMED, FIXED.** `Result` is `{ data?, error?: unknown, success }`. Two failures kept only the web-push error; two successes returned the web-push receipt and dropped FCM's `data`, so `messageId` named half the send. Successes now fold through `mergeSendResults` (`messageId`/`provider` joined, `response` keeps both bodies, `sent` is the conjunction, `timestamp` is the later of the two); two failures return an `AggregateError` carrying both causes; a mixed outcome still reports the failure. The misleading comment is gone — the fold's actual contract is documented on `mergeGroupResults`.

Tests (4 new, 124 total): both-succeed merge carries both message ids; mixed send with an unconfigured FCM channel throws having delivered nothing; a synchronously throwing transport still lets the other group through; both-fail keeps both causes. Verified they are real regression tests — restoring the previous send body fails exactly these four and nothing else.

Gates: 124 tests pass, `lint:types` clean, `lint:eslint` clean (the extraction of `mergeGroupResults` also keeps `send` under the cognitive-complexity limit), `api:check` 48/48 after a fresh build.

PR #456, opened 2026-08-21
perf(observability): prune buckets once per window

prisis:
## What was wrong

`recordFunctionMetric` and `recordAuthEvent` ran a retention `DELETE` — a correlated-subquery `MAX` scan plus a range delete — on **every** RPC dispatch and every auth attempt, the hottest write paths in the DO. Two sibling modules in the same package (`query-metrics.ts`, `metric-history.ts`) already established that this class of write is only needed once per minute bucket and gate it with a `WeakMap`-per-storage-handle marker; these two files had the same write shape but no gate.

## What it does now

Both recorders now carry the same marker: a module-level `WeakMap<object, number>` keyed by the storage handle (not a module scalar — workerd hosts several DO instances of one class per isolate, so a shared scalar would let a busy shard claim the window and starve every other shard's retention bound). The retention `DELETE` runs only when the current bucket differs from the last-pruned one for that handle.

Retention semantics are unchanged:
- The auth-metrics delete is table-wide, so once-per-window-per-handle is exact.
- The function-metrics delete stays per-`path`; a path not written in the winning window is pruned by its own next write in a later window — delayed by at most one window, never unbounded (stated in the code comment).

## Verification

- `pnpm --filter "@lunora/observability" run test` — 239 passed (237 existing + 2 new: one per recorder, asserting two records in the same bucket issue the prune `DELETE` once and a record in a later bucket issues it again)
- `pnpm --filter "@lunora/observability" run lint:types` — clean
- `pnpm --filter "@lunora/observability" run lint:eslint` — clean
- `pnpm run api:check` — all 48 snapshots match (no public surface change)


prisis:
Both review items addressed in `fb8b3e2`.

**Item 1 — confirmed, fixed.** Reproduced the reasoning against the code: `admitPath` gates first, then a per-handle marker claimed the window for all paths while the `DELETE` only covered `WHERE path = ?`. A path that never lands a window's first write was never pruned — genuinely unbounded, and my docstring's "delayed by at most one window" claim was wrong.

The marker is now keyed by `(handle, scope)` where scope is the row-group the caller's delete actually covers: `input.path` for function-metrics, the default unscoped key for auth-metrics' table-wide delete. Not the "winning call prunes all paths" option — a table-wide `bucket_ms <= MAX(bucket_ms) - retention` would evict an idle path's whole history whenever a busier path is newer, which the existing `trims only the path being written` test pins against. Key space is bounded by `FUNCTION_METRICS_MAX_PATHS` via the admission gate; noted in the code comment.

New test: `still prunes a second path written in a window another path claimed`. Verified it is a real regression test — temporarily restoring the per-handle marker fails that test and only that test (`1 failed | 239 passed`).

**Item 2 — agreed, extracted.** `src/prune-marker.ts` now owns the marker and its rationale; both recorders call `shouldPrune(sql, bucket, scope?)`. Internal module, not added to the curated barrel, so no public surface change. `query-metrics.ts` / `metric-history.ts` keep their own copies — out of scope here, and the shared helper is now the obvious landing spot when they are touched.

Gates: 240 tests pass (237 baseline + 3), `lint:types` and `lint:eslint` clean, `api:check` 48/48 after a fresh `build:packages`.

PR #457, opened 2026-08-21
fix(seed): respect unique columns when seeding

prisis:
## What was wrong

`@lunora/seed` generated each row's cells independently and never read the `unique` flag that `.unique()` places on column metadata. Seeding a table with a unique enum column at the default count of 10 was guaranteed to collide (`oneOf` over a 2-3 value domain), and a unique plain-string column near-certainly collided against faker's small lorem pool at higher counts. Since auth tables mark columns unique, this bit the standard stack — and the failure was a raw SQLite UNIQUE-constraint error with no attribution to the offending column.

## What it does now

- `FieldSpec` carries `unique` (read from the column metadata, mirroring how `nullable` is read).
- **Unique string-kind columns** derive row-uniqueness from the absolute row index: emails keep their address shape via a plus-tag (`local+3@domain`), everything else gets an `-<index>` suffix. Deterministic per (seed, table, index, column) like every other value, and safe across `indexOffset` batches.
- **Unique bounded-domain columns** (enum constraint, boolean) deal values without replacement, indexing a deterministic per-(table, column) shuffle of the domain. When the requested count exceeds the domain size, the plan fails fast *before generating* with a `BAD_REQUEST` naming the table, column, domain size, and requested count.
- **Non-unique columns** go through a completely untouched code path.

The studio's mirror generator has the same blindness; porting this there is deferred to a follow-up.

## Verification

- `pnpm --filter "@lunora/seed" run test` — 83 tests pass, including 4 new ones: 200 distinct values for unique string columns (deterministic across runs), a 3-member unique enum dealt without duplicates at count 3, the named fail-fast error at count 10, and a pinned pre-change output proving non-unique generation is byte-identical.
- All 79 pre-existing seed tests pass unchanged.
- `pnpm --filter "@lunora/seed" run lint:types` / `lint:eslint` — clean.
- `pnpm run api:check` — all 48 snapshots match (`FieldSpec` is internal; no public surface change).


prisis:
@coderabbitai /review

PR #476, opened 2026-08-24
feat(runtime): store the declared REST cache policy at the edge

prisis:
## Summary

Three independent caching/IO fixes that came out of evaluating whether Lunora should port something like [cachey](https://github.com/s2-streamstore/cachey) or [range-cache](https://github.com/xav-db/range-cache). The conclusion was no — the workload those Rust crates solve (repeated overlapping byte-range reads of immutable columnar blobs) does not exist here, and Workers have no disk tier — but auditing for it surfaced three real gaps. Each is a separate commit.

**1. `feat(runtime)` — `.expose({ rest: true, cache })` never actually stored anything** (`@lunora/runtime`, `@lunora/platform`)

`rest-cache` emitted `Cache-Control` / `Vary` and stopped there. A response a Worker *generates* is not stored by the colo cache on its own, so the declared policy bought browser revalidation and nothing else — every request still paid a shard dispatch. `caches.default` was used exactly zero times in the repo.

New `rest-edge-cache` adds the missing half: a `match` before dispatch, a `waitUntil`-deferred `put` after. Three guards make storing a procedure-backed response safe rather than a cross-user leak:

- Only a genuinely anonymous, effective-`public` exchange is stored, reusing the same credential check the header path applies. A declared-`private` policy is never stored — it is caller-specific by definition and this cache is shared by everyone hitting the colo.
- **`Vary` is enforced in the key, not delegated.** Cloudflare's cache honours `Vary` for `Accept-Encoding` only (as `shared/rest-surface`'s own docblock already noted), so a body varying on `x-lunora-shard-key` would otherwise be handed to a caller with a different key. Every varying header's value folds into the stored URL, turning the hazard into a miss.
- The lookup runs **after** the rate-limit gate — the order a CDN uses. A hit still costs a Worker invocation and is still a request that caller made, so it stays metered; it just skips the shard.

A cache read or write that rejects is treated as a miss, never a failed request. Hits carry `X-Lunora-Edge-Cache: hit`.

Per the platform-parity rule, `@lunora/platform` gains the `HttpCacheLike` contract and rates `httpCache` in both matrices: `native` on Cloudflare, `unsupported` on Node, where the surface degrades to emitting `Cache-Control` alone.

**2. `perf(storage)` — a ranged read opened a whole-object body and threw it away** (`@lunora/storage`, `@lunora/server`)

`serveStorageObject` started a full-object `download()` on every `Range` request purely to learn the object's size, cancelled that body, then issued the ranged read. Every video seek opened and discarded a whole-object transfer.

New `Storage.head(key)` — the body-free R2 read `getMetadata` already did internally, surfaced as the raw object shape so it carries the `etag` and base64 digest an HTTP response needs. The range resolves against that, then one ranged `download()`. Unranged requests skip the head entirely; an unsatisfiable range now needs no download at all.

Caught while wiring it: `storageRules`' allowlist would have dropped `head` — both an ungated route around `getMetadata`'s `read` rules and undefined under a guarded procedure. It is now gated as `read`.

**3. `refactor(shared)` — extract `memoizePromise`, fix two bugs in its copies** (`shared/`, `@lunora/mcp`, `@lunora/x402`)

Three lazily-built async singletons had each hand-rolled the same keyed memo. `shared/promise-memo.ts` is the one definition, and it fixes two bugs the copies had between them:

- `shared/hmac-url.ts` never evicted on rejection, so a single failed `crypto.subtle.importKey` poisoned that secret for the isolate's whole life.
- The two that did evict deleted whatever sat under the key at rejection time, not necessarily their own entry — a slow failure could delete a healthy retry that had replaced it. The shared helper compares identity first.

## Linked issues

None.

## Test plan

- [x] `pnpm run lint:types` passes (75 projects)
- [x] `pnpm run lint:eslint` passes (only pre-existing `css/use-baseline` warnings in `apps/docs/src/styles/app.css`)
- [x] `pnpm run lint:prettier` passes
- [x] `pnpm run lint:package-json` passes
- [x] `pnpm run api:check` passes — snapshots regenerated after a fresh build (`platform`, `runtime`, `storage`, `lunora`)
- [x] `pnpm run dist:check` passes after `pnpm run build:packages:prod`
- [x] `pnpm run test` — 4 failures across `@lunora/config`, `@lunora/studio`, `@lunora/hyperdrive`; **all pass in isolation**. Three are the documented parallel-run contention; the `config` one (`lint-ignores.test.ts`) is environmental — it `chmod 0500`s a directory and expects the write to fail, and this container runs as root. None of the three packages is touched by this PR.
- [x] New tests: 20 for the edge cache, 6 for `memoizePromise`, 4 pinning the storage round-trip count, 2 for `head`, 1 for `head` under storage rules

Reviewers may want to re-run `pnpm --filter "@lunora/{runtime,server,storage,platform,mcp,x402}" run test`.

## Checklist

- [x] Updated relevant docs in `apps/docs` (`concepts/rest-api.mdx`, `concepts/caching.mdx`, `concepts/file-storage.mdx`) and `packages/storage/README.md`
- [x] No `package.json` files in `packages/*` modified outside the touched package (only `packages/x402/tsconfig.json`, to drop `rootDir` for the new `shared/` import — the documented requirement for any `shared/` consumer)
- [x] No new packages added
- [x] Migration impact noted below

## Notes for reviewers

**Start with `packages/runtime/src/rest-edge-cache.ts`** — the cache-key construction is the part that carries the safety argument, and `restCacheKey`'s NUL-joined encoding is deliberate (a byte that cannot appear in a header field-value, so no combination of values can be made to collide).

Two behaviour changes worth flagging:

- **An app already declaring `.expose({ rest: true, cache: { scope: "public", … } })` will now have those responses stored at the colo**, where before they were not. That is the declared policy finally doing what it says, but it is a live change for existing deployments. `edgeCache: null` on `buildRestRoutes` is the opt-out.
- **`serveStorageObject`'s structural storage parameter now requires `head`** alongside `download` (marked `BREAKING CHANGE` in the commit). `@lunora/storage` provides it with its own fallback to a 0-length ranged `get()`; a hand-rolled test double must add it.

Two things I deliberately did **not** do:

- No request coalescing in the edge cache. It is cachey's other half, but the DO's `ReactiveCache` already memoizes by `(functionPath, argsHash)`, so the marginal win over isolate-level coalescing is small and the body-sharing hazards are not.
- `@lunora/server`'s `ReadOnlyStorage` type does not gain `head`. It already disagrees with `@lunora/storage`'s facade on `download`'s return type (a stream vs the metadata-rich object), so adding `head` there would not make `serveStorageObject(ctx)` typecheck and would just add an unmatched declaration. That mismatch is pre-existing and worth its own change.

https://claude.ai/code/session_01EjmVJXmP8D1Amh6t49vS4T

---
_Generated by [Claude Code](https://claude.ai/code/session_01EjmVJXmP8D1Amh6t49vS4T)_

## Summary by CodeRabbit

* **New Features**
  * Added edge caching for eligible public REST GET responses, with cache policies, request-aware privacy handling, cache-key variation, and cache status visibility.
  * Added storage `head()` support for retrieving object metadata without downloading the body.
  * Added HTTP cache capability reporting and configuration options for supported platforms.

* **Documentation**
  * Expanded caching, REST API, and file storage documentation with configuration, platform behavior, and usage details.

* **Bug Fixes**
  * Improved ranged storage responses by resolving metadata without an unnecessary full-object transfer.


prisis:
@coderabbitai /review

PR #492, opened 2026-08-26
fix(test): budget vitest workers against vis's task concurrency

prisis:
Independent of the perf stack, and worth landing on its own: **the repo's flaky tests are not flaky, they are starved.**

## The measurement

`vis` runs several package test tasks at once. Vitest, seeing a whole machine, spawns roughly one worker per core inside *each* of them. On a 10-core box that is five instances × ~9 workers — about **45 processes competing for 10 cores**.

The failures had the signature of starvation, not of a race:

- `findByTestId` calls that resolve instantly in isolation, missing a **1 s** budget by **3–6 seconds**
- a CLI test reported at **10,063 ms** against a **10,000 ms** limit

Every one passed on a re-run. Every one passed alone. That is exactly why they kept getting written off as noise — including by me, repeatedly, across this whole line of work.

## The fix

`maxWorkers` is now derived from **`VIS_TASK_SLOTS`**, which vis sets in every task it spawns to declare how many it runs concurrently.

Reading vis's own signal rather than mirroring `vis.config.ts`'s `parallel` means the two cannot drift. And a standalone `pnpm --filter <pkg> run test` sees no variable and keeps vitest's default — which matters: a blanket cap made a single-package run about **twice as slow** (shard-engine 9.4 s → 18.6 s), and that is the inner dev loop.

| | before | after |
|---|---|---|
| full suite | 3 failures across 3 runs | **5 consecutive clean runs** |
| wall time | ~290–380 s | **~190–200 s** |
| `--filter shard-engine` | 9.4 s | 7.4 s (unchanged by design) |

The contention was costing more than it bought — the suite is both reliable *and* faster.

## The timeouts were keyed backwards

Two places assumed CI is the contended environment:

```ts
testTimeout: process.env.CI ? 30_000 : 10_000,
configure({ asyncUtilTimeout: process.env["CI"] === "true" ? 5000 : 1000 });
```

The second even documented the reasoning — *"Local runs keep the snappy default for fast feedback."*

It is the opposite. CI gets a dedicated runner doing nothing else; `pnpm run test` locally fans 108 tasks across a machine that is also running an editor and a browser. **Local is the more contended environment and had the tighter fuse.**

Both are unconditional now. A timeout only bounds how long a *passing* assertion is willing to wait, so headroom costs nothing when the suite is healthy — and the alternative is a false failure that costs a three-minute re-run.

These are the backstop; the worker budget is the actual fix.

## Verification

Five consecutive `pnpm run test --no-cache` runs, 108/108 each. `lint:types`, `lint:eslint`, `lint:prettier` clean.

## Summary by CodeRabbit

* **Tests**
  * Improved test-runner resource allocation based on available task slots and CPU capacity.
  * Standardized test and hook timeouts to 30 seconds across all environments.
  * Preserved default worker behavior for standalone test runs.

### DIFF
--- a/packages/studio/__tests__/setup-reactflow.ts
+++ b/packages/studio/__tests__/setup-reactflow.ts
@@ -15,11 +15,20 @@ import { configure } from "@testing-library/react";
 /* eslint-disable vitest/require-hook -- a setupFile installs global stubs at module scope before any test runs; there is no hook context */
 
 // Lazy-loaded Studio panels (advisors, migrations, …) can take longer than
-// Testing Library's 1s default to mount under CI's parallel-test contention
-// (and v8-coverage instrumentation), so give `findBy`/`waitFor` more headroom
-// there — mirroring the repo's CI vitest-timeout bump. Local runs keep the
-// snappy default for fast feedback.
-configure({ asyncUtilTimeout: process.env["CI"] === "true" ? 5000 : 1000 });
+// Testing Library's 1s default to mount under parallel-test contention (and
+// v8-coverage instrumentation), so `findBy`/`waitFor` get more headroom.
+//
+// Deliberately NOT keyed on CI. The previous version gave 5s under CI and kept
+// "the snappy default for fast feedback" locally, which has it backwards: CI is
+// a dedicated runner, while `pnpm run test` locally fans 108 tasks across a
+// machine that is also running an editor and a browser. Local is the more
+// contended environment and had the tighter budget — these `findByTestId` calls
+// were failing at 4.2s and 6.5s against a 1s limit, on a machine where they pass
+// instantly in isolation.
+//
+// The timeout only bounds how long a PASSING query waits, so the headroom costs
+// nothing when the suite is healthy.
+configure({ asyncUtilTimeout: 5000 });
 
 /* eslint-disable class-methods-use-this -- no-op DOM stub: ResizeObserver methods intentionally do nothing */
 class ResizeObserverStub {
--- a/tools/get-vitest-config.ts
+++ b/tools/get-vitest-config.ts
@@ -1,5 +1,5 @@
 /// <reference types="vitest" />
-import { cpus } from "node:os";
+import { availableParallelism } from "node:os";
 
 import type { ViteUserConfig } from "vitest/config";
 import { configDefaults, coverageConfigDefaults, defineConfig } from "vitest/config";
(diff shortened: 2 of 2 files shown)


prisis:
Addressed — good catch, and it's the case that actually matters.

`cpus().length` and `availableParallelism()` agree on a developer machine and diverge exactly where this budget needs to be right: a CI runner is a CPU-limited container, so a 2-core executor on a 64-core host would have computed a cap of 12 and rebuilt the oversubscription the change exists to prevent.

Switched the divisor in `70e9580`. `availableParallelism()` is available on every Node this repo supports (`^22.15 || >=24.11`).

Re-verified after rebasing onto the alpha merge: `pnpm run test --no-cache` 108/108, types and lint clean.

prisis:
The two red checks here are not caused by this PR, and not flakes either.

`Test (node-22.15)` ran 55 tasks green and failed one: `search-core:test:coverage`. All 109 of its tests passed — it is the coverage threshold that fails, at lines 97.04% / statements 96.73% / branches 94.73%.

That gate is failing on `alpha` itself. I checked out a clean `alpha` and reproduced the same three numbers to the decimal. What this PR did was touch `tools/get-vitest-config.ts`, which makes every package affected, so `search-core`'s coverage task ran where it had been skipped as unaffected.

Fixed in #493 — an unreachable branch removed, and a reachable one given the test it was missing. Once that lands, a rerun here should be green.

prisis:
The third red check, CodSpeed, is also not this PR. It reports `Performance Regression: -3.9%` and attaches its own caveat:

> **Different runtime environments detected** — Some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results.

So the comparison it is unhappy about is one it already flags as unreliable.

Independently of that caveat, this branch cannot move a benchmark. Against `alpha` it changes exactly two files:

```
packages/studio/__tests__/setup-reactflow.ts | 19 +++---
tools/get-vitest-config.ts                   | 65 ++++++++++++++++---
```

Neither is reachable from a bench. No `vitest.bench.config.ts` imports `get-vitest-config` — benches use the separate `tools/get-bench-config.ts`, which has no `maxWorkers` and shares no code with it. And `setup-reactflow.ts` is a test-only setup file, referenced from neither `src/` nor `__bench__/`. Nothing here reaches a shipped bundle.

That accounts for all three failures on this PR: `Test (node-22.15)` was the pre-existing `search-core` coverage break (fixed in #493), `Check Test Run` is the aggregate rolling that up, and this one is cross-environment measurement noise.

PR #498, opened 2026-08-27
perf(storage): emit base64url in one pass instead of remapping base64

prisis:
Found by profiling `worker.fetch` — the entry every dispatched query and mutation goes through.

## The defect

`toBase64Url` built a binary string, handed it to `btoa`, then walked the result **three more times** to swap `+`/`/` and strip padding:

```ts
toBase64(bytes).replaceAll("+", "-").replaceAll("/", "_").replace(/=+$/, "")
```

Five passes and four intermediate strings, for what is one table lookup per 6 bits. Emitting the base64url alphabet directly skips all of it — and the padding is never produced rather than produced and then trimmed.

## Measured

Interleaved A/B against the previous implementation, median of 15 alternating slices, bundled from the actual shipped source:

| input | before | after | |
|---|---|---|---|
| 32 B (HMAC signature) | 661 ns | 182 ns | **3.63x** |
| 64 B (identity claims) | 1136 ns | 364 ns | **3.13x** |
| 256 B | 4091 ns | 1349 ns | **3.03x** |
| 4 KB | 59 µs | 20 µs | **2.90x** |
| 1 MB | 18 ms | 7.3 ms | **2.49x** |

Faster at every size, so there is no small/large trade to pick between. The small end is what matters: both callers live there and both are per-request — `encodeIdentityHeader` on every dispatched call carrying an identity, `signCanonical` on every signed-URL verify.

I am deliberately **not** quoting an end-to-end figure. Profiling put the encoder chain at ~4.8% of a Node `worker.fetch` run, but undici's `Request`/`Headers` dominate that profile and do not exist in workerd, so the Node end-to-end number understates it and is not worth reporting either way. A controlled A/B of the full path measured 18.8 → 18.0 µs; the first end-to-end reading I took was 35 µs and turned out to be load noise, which is why the table above is the isolated encoder instead.

## A duplicate deleted

`shared/hmac-url.ts` had hand-rolled its own `toBase64Url`/`fromBase64Url` — despite `shared/base64.ts` documenting itself as "the canonical base64url home rather than hand-rolling its own". It imports the shared one now.

Its `fromBase64Url` padded with `"===".slice((len + 3) % 4)` where the canonical one uses `"=".repeat((4 - (len % 4)) % 4)`. Equal for all four residues — checked before removing it, not assumed.

## Why the test is a golden rather than a round-trip

Output is byte-identical, fuzzed over every length from 0 to 400 (covering all three residue classes of the 3-byte group). That property matters more than the speed: **a signed URL is handed to a client and outlives the deploy**, so a signature the old encoder produced must still verify under the new one.

A round-trip test cannot see that break. If sign and verify both moved to a new encoding they would agree with each other and disagree only with every URL already in the wild. So the new test pins a literal golden signature — computed from the **old** implementation and then checked against the new one — plus the RFC 4648 vectors and the `-`/`_` alphabet swap.

Verified non-vacuous: swapping the alphabet back to `+`/`/` fails the golden, the alphabet assertion, and two pre-existing `create-storage` tests.

## Verification

`storage` 119 passed, `runtime` 950, `agent` 374, `bindings` 282. `api:check` clean across 54 snapshots, `lint:types` clean across 75 projects, eslint and prettier clean.

Note: `shared/` is in no package's vis cache key, so a cached build leaves every consumer's `dist/` on the old code. Everything here was built with `--no-cache`.

### DIFF
--- a/packages/storage/__tests__/base64url-stability.test.ts
+++ b/packages/storage/__tests__/base64url-stability.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from "vitest";
+
+import { fromBase64Url, toBase64Url } from "../../../shared/base64";
+import { signCanonical } from "../../../shared/hmac-url";
+
+/**
+ * `toBase64Url` is on two paths where the exact bytes are the contract, not an
+ * implementation detail:
+ *
+ * HMAC signatures for presigned/signed URLs: a URL handed to a client outlives
+ * any deploy, so a signature produced before a change must still verify after it.
+ * Round-trip tests cannot see a break here — if sign and verify both move to a
+ * new encoding they agree with each other and disagree with every URL already in
+ * the wild.
+ *
+ * The `x-lunora-identity` / `x-lunora-userid` headers, which cross the worker to
+ * shard boundary and can be produced and consumed by different builds during a
+ * rollout.
+ *
+ * So this pins literal expected output rather than asserting a round-trip.
+ */
+
+describe("base64url encoding is stable", () => {
+    it("encodes the RFC 4648 test vectors unpadded, with the URL-safe alphabet", () => {
+        expect.assertions(7);
+
+        const encode = (text: string) => toBase64Url(new TextEncoder().encode(text));
+
+        // RFC 4648 §10, minus the padding this variant does not emit.
+        expect(encode("")).toBe("");
+        expect(encode("f")).toBe("Zg");
+        expect(encode("fo")).toBe("Zm8");
+        expect(encode("foo")).toBe("Zm9v");
+        expect(encode("foob")).toBe("Zm9vYg");
+        expect(encode("fooba")).toBe("Zm9vYmE");
+        expect(encode("foobar")).toBe("Zm9vYmFy");
+    });
+
+    it("uses - and _ rather than + and /, for the bytes that produce them", () => {
+        expect.assertions(2);
+
+        // 0xFB 0xFF 0xFE is `+//+` in standard base64; every one of those four
+        // characters exercises the alphabet swap.
+        const encoded = toBase64Url(new Uint8Array([0xfb, 0xff, 0xfe]));
+
+        expect(encoded).toBe("-__-");
+        expect(encoded).not.toMatch(/[+/=]/);
+    });
+
+    it("round-trips every byte value and every 3-byte residue class", () => {
+        expect.assertions(3);
+
+        const all = new Uint8Array(256);
+
+        for (let index = 0; index < 256; index += 1) all[index] = index;
+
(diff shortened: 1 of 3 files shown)


prisis:
The red `SDK conformance (java)` check is not this PR, and it is not a conformance failure.

The job died at step 17, `Install google-java-format`:

```
curl: (22) The requested URL returned error: 429
##[error]Process completed with exit code 22
```

Maven Central rate-limited the runner. The pinned artifact is fine — this is not a checksum mismatch, and the SHA-256 line never ran. Steps 18-22, including **`Run the java conformance suite`**, were all `skipped`, so no conformance test executed at all.

This PR changes `shared/base64.ts`, `shared/hmac-url.ts`, and one storage test. It touches zero files under `sdks/`.

Queued a rerun for once the workflow finishes.

---

Worth a separate look: that step pulls an unauthenticated, uncached jar from `repo1.maven.org` on every run, so a 429 will take the java leg down again at random. `setup-java` can cache it, or it can go through `actions/cache` keyed on the pinned version — either removes the dependency on Maven Central being willing to serve a CI IP.

prisis:
Java conformance passed on rerun, confirming the 429 was transient. The remaining red check is CodSpeed: 5 improved, 1 regressed, 282 untouched.

**The regressed benchmark cannot reach this diff.**

It is `rank-page-hydrate :: in-batch: single IN(...) query + id->doc re-projection`. Within `shard-engine`, exactly one file imports `shared/base64` — `sql-projection.ts` — and it imports `toBase64`, the standard-alphabet function. This PR changes `toBase64Url` only; `toBase64` and `fromBase64` are untouched, which the diff shows directly:

```
-/** Base64 (from {@link toBase64}) -> URL-safe, unpadded base64url. */
-const toBase64Url = (bytes: Uint8Array): string => toBase64(bytes).replaceAll(...)
```

That is the only deletion in the file. Cursor encoding does not route here either — `query-args.ts` carries its own separate `toBase64`.

**On the run's variance.** One of the benchmarks CodSpeed lists as *improved* here is `ctx.metrics.count :: count, no attributes`. The same benchmark was reported as the *regression* on #492:

| PR | change under test | this bench |
|---|---|---|
| #492 | vitest worker config | 60.6 → 67.9 µs (**-10.67%**) |
| #498 | base64url encoder | 67.6 → 60.5 µs (**+11.80%**) |

Same benchmark, opposite directions, on two PRs that each touch something it cannot see — `packages/observability` reaches neither changed file. That is ±15% run-to-run noise on this suite, and CodSpeed flags the comparison itself as spanning different runtime environments.

So I would not read either the −16.79% or several of the +10-24% entries here as real. The measurement I trust for this change is the isolated interleaved A/B in the PR description: 2.49x-3.63x, median of 15 alternating slices, byte-identical output.

PR #503, opened 2026-08-27
fix(shard-engine): reject cursors minted before the tiebreak changed direction

prisis:
Acts on a review of the five changes merged in the last wave (#493, #494, #495, #498, #499). One real defect, two false comments, an incomplete dedup, and a review blind spot.

## The defect: #495 changed what existing cursors mean

Making the implicit `id` tiebreak follow the sort direction did not change a cursor's bytes — it changed their **meaning**. A cursor is an opaque `[...sortValues, id]` tuple with no version tag, so nothing could distinguish the two readings.

It is reachable. `use-paginated-core.ts` keeps page boundaries in `useState` for the component's life and replays `{cursor, endCursor}` on every reactive re-run, so a deploy feeds pre-change cursors to the changed predicate. Reproduced on six rows sharing one `_creationTime`:

```
current order:                          f,e,d,c,b,a
page with pre-#495 boundaries (b..d) →  [a]        (old semantics: [c,d])
```

Not empty — it returns a row from the *previous* page and makes four rows unreachable, silently, shaped exactly like a correct page.

Cursors now carry a two-character marker outside the base64 alphabet; one without it is refused as the existing typed 400. A client resets the feed and recovers. Wrong rows presented as right ones do not.

**Three mint sites feed one reader** and only the first was obvious — the DO's row cursors, shard-engine's rank cursors, and `sql-store`'s rank cursors. The last two surfaced as test failures, which is why the marker lives on the package index rather than in one module.

Credit where due: CodeRabbit flagged this risk on #495 (*"cursors created before deployment may be incompatible during mixed-version rollouts"*) and it was merged without a response.

## Two comments asserted things that are false

The codegen bench claimed `beforeAll` runs in a different context from bench bodies under CodSpeed's instrumented runner. **It does not** — `AnalysisRunner` calls the suite's `beforeAll`/`afterAll` around the benchmarks in the same process, which I verified in the plugin source and with a probe. Left standing, that claim implies the 28 bench files using `beforeAll` are inert. They are fine. The lazy priming stays (it is still the better shape); its justification is now accurate.

`ctx-db.ts` still documented an `id ASC` tiebreak, and `ctx-db-migrations.ts` still described the mixed-direction shape as current — in the docblock that justifies the index sort keys.

## The dedup was one of four

`shared/base64.ts` calls itself the canonical base64url home. Three copies were still hand-rolling it (`connector-cdc`, `scheduler-do`, `notify/web`), and a storage bench inlined the exact decoder #498 deleted under a comment claiming it mirrors verify. All four now use the shared one.

## `shared/` was invisible to review

It matched no path filter in `.coderabbit.yaml`, so #494 and #498 were skipped in full — the tree holding the cache-key encoder and the signed-URL codec. One line.

## Smaller

- `tiebreakDirectionFor` reads one field off the last key, so demanding a full `OrderKey[]` made the `sql-store` caller allocate N objects to inspect one string. Widened; adapter deleted.
- The team-chat assertion sorted with `localeCompare`, depending on the runner's ICU build — the exact locale sensitivity the key encoder documents avoiding. Set comparison instead.
- `broadcastDelta`'s comment pinned numbers measured against a fake whose `deserializeAttachment` is a field read, where workerd does a structured-clone deserialize per socket. Numbers cut; the qualitative claim stands.

## Measured and not taken

A single-pass `fromBase64Url` to match the encoder: **1.61x faster at 32 bytes, but 1.14x and 1.77x slower at 64 bytes and 4 KB.** Native `atob` beats a JS loop once inputs grow — the opposite of the encode side, where three `replaceAll` passes were the cost. Reverted.

An earlier draft of it also *skipped* invalid characters where `atob` throws. `decodeUserIdHeader` turns a throw into "anonymous", so that would have let a malformed header decode to garbage bytes and yield a userId nobody sent. An existing test caught it — worth recording as the reason that function must keep throwing.

## Verification

Full suite `--no-cache`: **108/108 task groups green**. `api:check` clean across 54 snapshots (the two additions are in the diff), `lint:types` 75 projects, `lint:eslint` 58 projects, prettier clean.

The cursor-rejection test was checked for vacuity: a legacy cursor alone rejects even without the marker check (slicing unaligned base64 happens to break `JSON.parse`), so the test also feeds a real payload behind a *wrong* marker — which does pass through without the check.

## Summary by CodeRabbit

* **New Features**
  * Added versioning to keyset cursors with a `~2` prefix.
  * Invalid or unrecognized cursor formats are now rejected with a clear bad-request error.
  * Exposed the cursor version prefix for integrations that need to identify supported cursor formats.

* **Compatibility**
  * Existing unprefixed cursors are no longer accepted and may need to be regenerated.

* **Documentation**
  * Clarified sorting and cursor behavior for descending queries.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -25,6 +25,10 @@
 
     "path_filters":
         - "packages/**"
+        # `shared/` is bundler-inlined rather than a package, so it matched no
+        # filter here and was skipped in full — including the cache-key encoder
+        # and the base64url codec on the HMAC signed-URL path.
+        - "shared/**"
         - "apps/**"
         - "tools/**"
         - "scripts/**"
--- a/api-snapshots/shard-engine.api.md
+++ b/api-snapshots/shard-engine.api.md
@@ -320,6 +320,12 @@ const COMMIT_SEQ_FIELD = "_commitSeq";
 const COMMIT_SEQ_TABLE = "__commit_seq";
 ```
 
+### `CURSOR_PREFIX` (const)
+
+```ts
+const CURSOR_PREFIX = "~2";
+```
+
 ### `CacheEntry` (interface)
 
 ```ts
@@ -4645,7 +4651,9 @@ const throwingScheduler: SchedulerLike;
 ### `tiebreakDirectionFor` (const)
 
 ```ts
-const tiebreakDirectionFor: (keys: ReadonlyArray<OrderKey>) => SortDirection;
+const tiebreakDirectionFor: (keys: ReadonlyArray<{
+    direction?: string;
+}>) => SortDirection;
 ```
 
 ### `trimCdcChanges` (const)
--- a/examples/team-chat/__tests__/chat.test.ts
+++ b/examples/team-chat/__tests__/chat.test.ts
@@ -65,11 +65,10 @@ it("posts messages into a channel and reads them back", async () => {
     // disagree. A table that needs write order declares `.commitOrdered()` and
     // reads `orderBy: [{ _commitSeq: "asc" }]`; two chat messages a millisecond
     // apart do not need it.
-    expect(messages).toHaveLength(2);
-    expect(messages.map((message) => `${message.authorId}:${message.content}`).toSorted((left, right) => left.localeCompare(right))).toStrictEqual([
-        "u-ada:hello",
-        "u-grace:hi back",
-    ]);
+    // Compared as a set, with no comparator: `localeCompare` would make this
+    // assertion depend on the runner's ICU build, which is exactly the locale
+    // sensitivity the shared key encoder documents avoiding.
+    expect(new Set(messages.map((message) => `${message.authorId}:${message.content}`))).toStrictEqual(new Set(["u-ada:hello", "u-grace:hi back"]));
 });
 
 it("keeps channels apart", async () => {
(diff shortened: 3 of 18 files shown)


prisis:
CodSpeed reports `normalize-log-fields :: flat 3 primitives` at -10.84%. That benchmark lives in `packages/do`, and this PR's entire `packages/do` diff is comments.

Proven rather than eyeballed — transpiling both sides with esbuild (which strips comments) gives byte-identical output:

```
$ cmp -s sd-alpha.js sd-next.js && echo IDENTICAL
IDENTICAL after transpile — the packages/do diff is comment-only
```

So there is no executable change in that package for a benchmark to observe.

## This is the third consecutive PR

| PR | change under test | benchmark flagged | delta |
| --- | --- | --- | --- |
| #492 | vitest worker config | `metrics.count :: no attributes` | **-10.67%** |
| #498 | base64url encoder | `rank-page-hydrate :: in-batch` | **-16.79%** |
| #503 | comments + cursor marker | `normalize-log-fields :: flat 3` | **-10.84%** |

None of the three benchmarks can reach the diff it was flagged against. #498 is the clearest single data point: the benchmark that regressed **-10.67%** on #492 was reported as *improved* **+11.80%** on #498 — same benchmark, opposite directions, on two PRs that each touch something it cannot see.

That puts the run-to-run spread at roughly ±15%, which is above the threshold the gate fires at. The consequence is what we are seeing: a perf gate that reports a red regression on most PRs regardless of content, which trains everyone to ignore it — including on the day it is right.

Worth either widening the threshold past the observed noise floor, or pinning the runner environment (CodSpeed itself attaches a "different runtime environments detected" caveat to these comparisons). Not something I am going to change unilaterally, but it is now well-evidenced.

PR #508, opened 2026-08-27
feat(studio): inline AI rewrite for the SQL console

prisis:
> **Draft — this does not pass CI.** Recovered from uncommitted work left in a stale worktree and
> replayed onto current `alpha` so it stops living on one machine's disk. Opened as a draft to make
> the remaining work visible, not because it is ready.

## What

The pieces the SQL console needs to propose an AI rewrite of the current statement in place:

- `line-diff.ts` — a line-level unified diff.
- `sql-inline-edit.tsx` — the inline edit surface that renders it.
- `assistant-reason.ts` — operator-facing copy for a degraded `aiGenerateSql` reply.

## Decisions worth keeping

- The diff is a hand-rolled LCS rather than a dependency: the engine caps what it rewrites at 2,000
  characters, so the quadratic table is a few hundred cells and cannot grow without bound.
- The table is walked forward from the origin, which keeps a run of removals grouped ahead of the
  matching additions instead of interleaving them line by line — the difference between a diff that
  reads like a change and one that reads like noise.
- Unified `<ul>` rather than side-by-side: the editor is a plain textarea with no diff view to
  borrow, the console has no width for two panes above a full-height editor, and unified survives a
  screen reader. Colour is never the only signal — every row carries a gutter marker.

## What is left to do

1. **Wire it.** Nothing imports `sql-inline-edit`, so the component is unreachable from the running
   console. The four failing tests are all this: they render the real editor pane and press
   `Ctrl+I`, and nothing handles the key. The one test that passes —
   *"does not arm on a deployment that cannot run the assistant"* — passes vacuously, since nothing
   arms under any condition.
2. **Register the i18n keys.** `lint:types` fails with three TS2345s: `"Proposed edit"`, `"Accept"`
   and `"Reject"` are not in the translation key union.
3. **Clear 16 eslint errors**, including a `||` that should be `??` (line 184) and an
   `no-unsafe-argument` on the degraded-reason path (line 231).
4. **Deduplicate the degraded-reply copy.** `sql-assistant-bar` still inlines its own rather than
   calling the shared `assistant-reason` — the sharing that module's doc comment claims has not
   happened yet.

`line-diff.test.ts` passes on its own and is the part of this worth trusting today.


packages/studio/src/features/sql/sql-inline-edit.tsx
```diff
@@ -0,0 +1,238 @@
...
+    const t = useT();
+
+    return (
+        <ul
+            aria-label={t("Proposed edit")}
+            className={cn("max-h-48 overflow-auto rounded-md border border-border bg-background", EDITOR_TEXT_CLASS)}
+            data-testid="sql-inline-diff"
+        >
+            {lines.map((line, index) => (
+                <li className={cn("flex gap-2", ROW_CLASS[line.kind])} data-kind={line.kind} key={`${index.toString()}-${line.text}`}>
```
prisis, line 57: Verified against the component; keeping the index-bearing key, because the rule's premise does not hold for this list.

**There is no stable id to use.** A `DiffLine` is `{ kind, text }` (`line-diff.ts`) emitted by the LCS walk — no identity of its own. Two identical lines in one statement (a bare `)`, a blank line, a repeated `AND …`) are genuinely indistinguishable, so a content-only `key={`${kind}-${text}`}` would collide and React would warn about duplicate keys. The current `key={`${index}-${text}`}` is always unique. Minting an id inside `lineDiff` would be the index under another name, and `crypto.randomUUID()` in render would remount every row on every render — both worse than the index.

**The list cannot reorder, sort or filter.** `lines` is re-derived whole from `(source, proposal)`; the `<ul>` renders only while `proposal !== null`; and `proposal` only ever goes `null → value`, because the Rewrite control exists solely in the `proposal === null` branch — a second attempt has to go through Reject, which sets `proposal` back to `null` and *unmounts* the list rather than swapping its rows. So the list appears on mount and disappears on unmount; it is never diffed against a different list.

**No per-row state to misattribute.** Each `<li>` is three static spans (the `+`/`-` marker, the visually-hidden Added/Removed word, the line text) — nothing focusable, no input, no selection, no open editor. The harm the rule names (a row's state following the wrong item across a reorder) needs both a reorder *and* per-row state, and this list has neither. Worst case is a wasted DOM re-create on a list the engine caps at 2,000 characters of SQL.

Same reasoning as the existing precedent one file over: `sql-editor-pane.tsx:82` keys the gutter line numbers by index.

No disable comment added — the repo forbids silencing a rule to go green, so this warning is expected to stay. React Doctor is green either way (score 89, 0 errors, 1 warning).

I did write this rationale into the file, then had to drop it: the pre-commit ESLint gate rejects **any** commit touching `sql-inline-edit.tsx` on this branch. The slice is missing `src/hooks/use-assistant-rpc.ts`, so line 5's import is unresolvable and typed linting cascades into 16 `no-unsafe-*` errors — the same root cause as the red `Lint (eslint)`, `Lint (types)` and both `Test` jobs here, and why `SqlEditorPanel` still has no `SqlInlineEdit` wiring (the 4 failing cases in `sql-inline-edit.test.tsx` arm ⌘/Ctrl+I on a panel that does not render it yet; `line-diff.test.ts` passes, 5/5). Worth pasting this above the `lines.map(` once that lands:

```tsx
{/* Keyed by position, not identity: a diff row is a kind plus a line of
    text with no id of its own, and two identical lines in one statement
    are indistinguishable — so the content alone cannot key the list.
    Position is sound here because the list never reorders, sorts or
    filters: it is re-derived whole from source and proposal, and Reject
    unmounts it rather than swapping its contents. The rows hold no state
    either — three static spans, nothing focusable — so there is nothing a
    remount could carry over to the wrong row. react-doctor's
    no-array-index-as-key needs both a reorder and per-row state to bite,
    and this list has neither. */}
```

prisis:
## Made this branch stand on its own

`SqlInlineEdit` had no call site and imported `../../hooks/use-assistant-rpc`, which exists only on `feat/studio-ai-chat`. The unresolved type degraded to `any` and cascaded into ~16 `no-unsafe-*` errors; with `--max-warnings=0` the pre-commit hook rejected any commit touching the file, and the five tests were 4 failed / 1 passed. All of that is fixed.

### Decision: **Option B — narrow the contract.** Measured closure that decided it

Porting `use-assistant-rpc.ts` verbatim (Option A) would have dragged in, none of which exists on this branch:

| Missing on `feat/studio-sql-inline-edit` | Backed by |
| --- | --- |
| `client.streamRpc` | `packages/client/src/lunora-client.ts` (+103), `http-stream.ts`, new `shared/sse.ts` |
| `ADMIN_FUNCTIONS.aiChat` / `aiNameQuery` / `aiCronExpression` | `packages/runtime/src/ai-chat-rpc.ts` (new, 371 lines), `create-worker.ts` (+130), `packages/do/src/shard-do.ts` (+82), `packages/server/src/http.ts`, `shared/ai-chat.ts` (new, 1005 lines), `shared/cron-expression.ts` (new) |
| 10 of 17 admin types (`SchemaFact`, `ChatTurn`, `ChatApproval`, `ChatStreamEvent`, `ChatResult`, `ChatToolReport`, `ChatPendingApproval`, `AiOptInLevel`, `GenerateCronResult`, `GenerateQueryNameResult`) | the above |

That is ~2.5k lines of #429's own feature surface, not a prerequisite of this one — squarely the "substantial chat/cron machinery plus server-side admin functions that do not exist on this branch" case, so Option B.

Option B here needed **no new abstraction at all**: this branch already has `useSqlAssistant` / `SqlAssistant`, the direct ancestor that `feat/studio-ai-chat` renames to `useAssistantRpc` / `AssistantRpc`. `SqlInlineEdit` reads exactly four members — `unavailable`, `pending("sql")`, `reason("sql")`, `rewrite` — and `SqlAssistant` already had three. So `rewrite` was added to it (via the same `askForSql` extraction #429 uses), and the component's prop is typed `SqlAssistant`. #429's `AssistantRpc` satisfies that structurally, and #429 deletes `use-sql-assistant.ts` outright, so the later merge is a type swap in one file with nothing to reconcile.

### Ported, and whether byte-identical

| File | From `feat/studio-ai-chat` | Byte-identical |
| --- | --- | --- |
| `hooks/use-sql-editor-surface.tsx` | whole file | ✅ — its entire diff there *is* the inline-edit slice (chord, armed span, accept/close, the keydown split that makes room for a third chord) |
| `__tests__/features/sql/sql-inline-edit.test.tsx` | whole file | ✅ — restores the one test the slice had dropped ("drops an armed rewrite when the draft is edited under it") |
| `line-diff.ts`, `line-diff.test.ts`, `assistant-reason.ts` | already on this branch | ✅ (verified, unchanged) |
| `sql-editor-panel.tsx` | inline-edit slice only | partial by design — `inlineEditEnabled`, `closeInlineEdit` on tab switch, three props. No chat, explain, seeded questions, rename, or `assistant`→`rpc` rename |
| `sql-editor-pane.tsx` | inline-edit slice only | partial by design — same, keeping the existing `assistant` prop name |
| `sql-inline-edit.tsx` | already on this branch | 2 lines differ (import + prop type) + the key comment below |
| `packages/do/src/sql-assistant.ts` | `editSql` handling | prompt clause + `GenerateSqlArgs` docblock + test byte-identical |

### The gap that made the feature inert

`editSql` had **no server-side support** anywhere on this branch — the component sent it and `aiGenerateSql` ignored it, so every "rewrite" was silently a fresh draft. Added to `packages/do/src/sql-assistant.ts`: the field, its cap, and its own prompt clause, deliberately separate from `failedSql` (a repair prompt applied to a working statement invites the model to invent a fault it can be seen to fix). Test ported too.

### Also

- **i18n**: 9 inline-edit ids added with #429's exact spelling, plus the 3 degraded-reason strings. `sql-assistant-bar.tsx` now renders the shared `assistantReasonMessage` instead of its own untranslated copy — that is the change that makes `assistant-reason.ts` the shared module its docblock already claimed, and it is the only non-inline-edit part of #429's bar diff.
- **`react-doctor/no-array-index-as-key`** (resolved thread, `sql-inline-edit.tsx:45`): the rationale is now recorded in a comment at the keyed element. No `eslint-disable`. react-doctor 0.9.11 reports **zero** findings anywhere under `features/sql/**`.

### Gates

| Gate | Result |
| --- | --- |
| `pnpm run lint:eslint` (repo-wide) | ✅ **0** — 58 projects, 161 tasks |
| `pnpm run lint:types` (repo-wide) | ✅ **0** — 75 projects |
| `pnpm run lint:prettier` | ✅ |
| `pnpm run lint:package-json` | ✅ 76/76 sorted (no manifest touched) |
| `@lunora/studio` tests (`unit` + `component`) | ✅ 144 files / 1289 tests; the 6 inline-edit tests pass for real, nothing skipped or weakened |
| `@lunora/do` tests | ✅ 61 files / 711 tests |
| `pnpm --filter "@lunora/studio" run build` + `build:prod` | ✅ — meaningful now that `SqlInlineEdit` is in the bundle graph |
| `pnpm run build:packages` | ✅ 54 projects |
| `pnpm run api:check` | ✅ 54/54 snapshots match (`GenerateSqlArgs` is not re-exported from `@lunora/do`'s entry, so `editSql` does not move the surface) |
| `react-doctor` 0.9.11 (pinned, not bumped) | ✅ no new findings |
| `pnpm run dist:check` | ❌ **pre-existing** — verified red on this branch with my changes stashed. `build-standalone.mjs` falls back to `NODE_ENV=development` whenever anything still references `jsxDEV`; `feat/studio-ai-chat` is what fixes it (`needs-dev-jsx-runtime.mjs` + `check-dist-production.js`). Out of scope here. |
| `pnpm run test:workerd` | ⚠️ **could not run** — workerd refuses to boot in this sandbox (`No such module "cloudflare:test-internal"`). Nothing here touches SQL execution, storage or DO state: the `@lunora/do` change is prompt-string assembly only, covered by the node-project test added alongside it. |

Left as a draft.

packages/studio/src/features/sql/sql-inline-edit.tsx
```diff
@@ -0,0 +1,256 @@
...
+                a repeated `AND ...`), so a content-only key collides. The three
+                conditions the rule exists to protect are all absent: the list cannot
+                reorder — it renders only while `proposal !== null`, and `lineDiff`
+                recomputes the whole array from a new proposal; a retry goes through
+                Reject, which nulls `proposal` and UNMOUNTS the list rather than
+                re-ordering it; and no row holds per-row state or focus that a
+                mis-keyed re-render could carry onto the wrong line.
+            */}
+            {lines.map((line, index) => (
+                <li className={cn("flex gap-2", ROW_CLASS[line.kind])} data-kind={line.kind} key={`${index.toString()}-${line.text}`}>
```
prisis: Fixed in a978209ca — no index key and no suppression. `lineDiff` now gives every row its 1-based line number on each side it exists on (`before` for removals, `after` for additions, both for context), and `DiffView` keys by `${before}:${after}`. That pair is unique per row even when the text repeats (`)`, a blank line), so it is a real identity rather than the array position. The long false-positive comment is gone.

PR #545, opened 2026-09-01
docs(server): scope the mutation atomicity claim to shard-local writes

prisis:
## What

The server package docs claimed every `ctx.db` write in a mutation commits atomically and rolls back together. That is true only for tables that live in the shard, which share one Durable Object storage transaction. A `.global()` table lives in D1, which has no interactive transactions, so a mutation writing a shard-local row and a global row writes them separately; a throw after the global write leaves it in place. The scaling tutorial already said this; the server page contradicted it.

## Change

- Scope the claim to shard-local writes and say why (one DO storage transaction).
- State the `.global()` boundary with a link to the scaling tutorial, and what to do about it: order the global write last, or move the pair into a workflow step whose replay is idempotent.

Docs only. Source file is `packages/server/docs/index.mdx` (the `apps/docs` copy is generated and gitignored).

Stacked on #544.


prisis:
@coderabbitai review

prisis:
@coderabbitai full review

PR #547, opened 2026-09-02
fix(payment): record facade refunds and key them by refund id

prisis:
## What

A refund issued through the facade wrote only `state: "refunded"` and left the total to the webhook, but the state machine had no transition out of `refunded`, so the provider's confirmation was rejected as `illegal_transition`. The ledger stayed at zero, the over-refund guard read that zero, and a second refund reached the provider. Verified by execution.

## Changes

- The facade records the amount it issued; `refunded` accepts the provider's confirmation idempotently (absolute vs delta per provider).
- A refund arriving before its capture is `orphaned` and retried once, like `subscription.updated`, instead of burning its event id.
- Polar `reportUsage` passes the idempotency key as `externalId`.
- `refundPayment` returns the provider refund id; the local marker is keyed on it so two same-amount refunds count once each.

Chain 2 of 8, stacked on #546.

---

Every branch in this chain was verified on its own: `build:packages`, `api:update`/`api:check`, golden-fixture capture, `lint:generated`, `vis run lint:types --no-cache`, and the group's package tests, all exit 0. The full gate suite (eslint, prettier, package-json, whole-repo no-cache tests, prod build + `dist:check`) is green on the chain tip.

## Summary by CodeRabbit

- **New Features**
  - Refund results now include optional provider refund identifiers and pending status for easier tracking.
  - Refund processing supports partial refunds and records eligible refunded amounts immediately.
  - Repeated refund requests are handled safely without exceeding the captured amount.

- **Bug Fixes**
  - Prevented duplicate refund counting when provider confirmations arrive.
  - Improved handling of refunds received before payment capture, with bounded automatic retry support.
  - Improved usage-event deduplication during retried processing.


prisis:

---

PR #554, opened 2026-09-02
fix(client): hand the read cache to the subscription that is already open

prisis:
## What

The durable read cache is loaded after a microtask and an async adapter, while every framework adapter's first `subscribe` is synchronous at mount. The cache entry therefore missed the live subscription, stayed parked, and was consumed by the NEXT subscribe of the same key (navigate away and back, or a React remount), replaying the previous session's value synchronously: deleted rows flashed back and new rows vanished until the server snapshot landed. Verified by execution.

## Changes

- The cache load hands each entry to the subscription already open for its key (no-op once a live value exists) and drops it; only keys with no live subscription are held for a later subscribe.
- `hydrateOnStart` docs say what the code does: only React ever waited on `whenReady`, and the wait is no longer needed.
- React: paginated hooks gain `onError` and `error` and no longer stick at `LoadingMore` after a failed page; `usePreloadedQuery` takes `onError`; `useSubscription` keeps the server `code`; `useAgent`/`useAgentChat`/`usePresence` expose `error`.
- The registry's polling fallback (which only ever fired for a closed client or unencodable args, then hid the error in a 5 s loop) is removed; the throw surfaces.
- `subscribe()` docblock matches the body (a follower tab may subscribe; that is the relay).

Chain 9, stacked on #553.

---

Verified on its own like the round-7 chain: `build:packages`, `api:update`/`api:check`, golden-fixture capture, `lint:generated`, `vis run lint:types --no-cache`, and the group's package tests, all exit 0. Every fix has a regression test that was shown red on the unfixed code.


prisis:

---

PR #556, opened 2026-09-02
fix(templates): stop the scaffolds shipping a 403 home page and a split-brain write

prisis:
## What

Three commits from a hunt over storage, queue, access and project scaffolding:

- **templates** — the Next.js scaffold preloaded its home page with a `shardKey` while its app allowed no unauthenticated shard access, so the first load was a 403; its send button wrote to the root shard while the feed read the channel shard. Every other template had the flag. A static check in `tests/vis-templates` now requires any template that sends a `shardKey` to declare a shard gate. The default `lunora init` overlay scaffold is now byte-identical to the standalone template (enforced `.max()`, an `ip` fallback on the limiter key, the rate-limit schema actually used). The create-vite base is copied without symlinks and pinned to a SHA. Registry object roots merge key-wise instead of replacing a whole `vars` block. The expo template folds `?token=` into the auth header only on the WebSocket upgrade.
- **dispatch** — a rotated or missing admin token made the scheduler dispatch route answer `FORBIDDEN` 403, which the queue runner classified as per-message poison and acked, draining the queue one message per delivery. The bearer check now throws `DISPATCH_UNAUTHENTICATED`, which the runner treats as infrastructure and never acks.
- **server** — `serveStorageObject` takes a required fail-closed `authorize` seam (docs and the storage skill claimed it verified signed URLs; it did not), always sends `nosniff`, and serves non-media as an attachment; storage rule prefixes match on a path segment (`users/1` no longer grants `users/10/…`); `Storage` exposes its `bucketName` so a hand-wired named bucket signs and matches rules consistently.

BREAKING: `serveStorageObject` gains a required options argument; `Storage.bucketName` is new; new catalog code `DISPATCH_UNAUTHENTICATED`.

Chain 11, stacked on #555.

---

Verified on its own like the rest of the chain: `build:packages`, `api:update`/`api:check`, golden-fixture capture, `lint:generated`, `vis run lint:types --no-cache`, and the group's package tests, all exit 0. Every fix has a regression test that was shown red on the unfixed code; the Next.js template additionally passed the scaffold/install/build/typecheck smoke.


prisis:

---

PR #562, opened 2026-09-02
fix(observability): close the four span and log bounds the docs claimed were held

prisis:
## What

Eleven findings in observability and the runtime sinks.

- **A negative batch cap hung the isolate** on its first telemetry event, and a zero or fractional one discarded everything; the sink now refuses the value at construction.
- **OTLP logs shipped raw fields and messages** while the console and span pipelines redact by default, so `ctx.log.info("charged", { email })` was masked on one sink and cleartext at the collector. Redaction is now the default there too, with a documented opt-out.
- **A refusing collector was indistinguishable from success**: every non-OK response was ignored silently. It now reports status and host through the rate-limited channel the tail sampler already uses.
- **The security audit warned of an admin exposure that does not exist** — `LUNORA_WS_BEARER` gates ordinary user subscriptions, not admin ones — and pointed the operator at the wrong remediation. The finding now says what the variable does, and its level drops to warning accordingly.
- Plus: the 128-key span-attribute bound was bypassed by two writers, forward log paging skipped rows, the request-log table was re-ensured on every append, and four docblocks disagreed with the code about sampling keys, tail-sampler scope, series eviction and security-header precedence.

Chain 17, stacked on the branch below.

---

Verified on its own like the rest of the chain: `build:packages`, `api:update`/`api:check`, golden-fixture capture, `lint:generated`, `vis run lint:types --no-cache`, and the group's package tests, all exit 0. Every code fix has a regression test that was shown red on the unfixed code.


prisis:

---

PR #578, opened 2026-09-02
fix(server): scope a deferred schedule to its own transaction

prisis:
The deferred-schedule buffer hung off the facade, so every transaction reachable through a context wrote into one shared list and a settle returned early while any window was still open.

- **The buffer is now one window per transaction.** A buffered call goes to the window innermost at call time; settle removes its window by identity, since windows settle in whatever order their transactions resolve rather than the order they opened. A rollback drops only its own jobs, and a commit hands them to the nearest still-open ancestor — SQLite in a Durable Object has no savepoints, so a nested span cannot commit, and must not schedule, alone. This ends both halves of the shared-buffer defect: a sibling's commit no longer dispatches a rolled-back window's jobs, and a committed window no longer loses its own to another window's rollback.
- **`runAt` had no bound at all**, so a non-finite instant reached the Durable Object, rendered as `null`, and produced a record no alarm can fire. It now converts to a delay and goes through the same shared guard `runAfter` uses. A past instant is still accepted and clamped — `runAfter(0, …)` reaches `runAt` a fraction of a millisecond after reading its own clock, and `runAt(row.dueAt, …)` on an overdue row is legitimate.
- **A mid-flush dispatch failure abandoned the rest of the queue** and skipped the deferred-delete flush that follows it. The flush now drains everything and reports afterwards, with the delete flush in a `finally`. A lone failure is rethrown as itself so the scheduler's coded refusals survive; several become one `AggregateError`.
- **The deferral was installed on fewer paths than the transaction was.** A `stream` context, and any context with no registered kind, got the BEGIN/COMMIT span for a composed mutation with the buffer missing. `createCaller` was worse than filed: it invoked handlers directly with no span at all, so from an action or stream context it got neither. Both now route through the one place the span, the deferral and the delete flush live.
- **A caller's own job id was discarded** inside a mutation and replaced with a generated UUID, while being honoured outside one. The facade now applies the exact rule the Durable Object applies, so a caller-supplied id reaches the duplicate-id refusal.
- **The scheduler's Durable Object client flattened every non-2xx to `INTERNAL`**, whose message is redacted — so the new `409 DUPLICATE_SCHEDULE_ID` reached developers as "Internal error". Coded envelopes are now propagated; genuine transport failures stay `INTERNAL`.

**Behaviour change worth a reviewer's eye:** a nested window reporting a rollback now drops its own jobs rather than leaving them to the enclosing span. Without savepoints its writes may still land if the outer transaction catches and commits, so this is a trade rather than a strict win — "the mutation did not complete, so do not run its after-commit job" is the fail-safe direction.

**Not fixed, deliberately:** the deferred-*delete* facade has the same narrow gate, but widening it without adding a flush to the stream and subscription paths would swap a loud `TypeError` for a silent leak. Filed rather than half-done.

## Also here: the `search-core` coverage gate

`search-core:test:coverage` fails its own declared thresholds and turns every CI run red once it gets that far. It **predates this stack** — the same numbers reproduce on the base — and stayed hidden because the audit chain's earlier PRs failed that job earlier, in the affected-set step, and because `vis run test` does not check coverage at all.

Three gaps, all reachable through the public surface: `searchIndexField` had no test (the one uncovered function), the CJK expander was never handed a run holding no CJK, and no text put a Latin or digit stretch *after* a CJK run rather than before it. Coverage goes to 100% lines, 100% statements and 97.14% branches against thresholds of 100 / 99 / 95.

---

### DIFF
--- a/api-snapshots/scheduler.api.md
+++ b/api-snapshots/scheduler.api.md
@@ -508,7 +508,13 @@ interface WorkpoolOptions extends LunoraSchedulerOptions {
 ### `assertScheduleDelay` (const)
 
 ```ts
-const assertScheduleDelay: (delayMs: number, surface: string) => void;
+const assertScheduleDelay: (delayMs: number, surface: string, argument?: string) => void;
+```
+
+### `assertScheduleInstant` (const)
+
+```ts
+const assertScheduleInstant: (timestampMs: number, nowMs: number, surface: string) => void;
 ```
 
 ### `assertValidCronExpression` (const)
@@ -581,6 +587,12 @@ const isValidCronExpression: (schedule: string) => boolean;
 
 ```ts
 const isWorkflowReference: (target: unknown) => target is WorkflowReference;
+```
+
+### `resolveScheduleId` (const)
+
+```ts
+const resolveScheduleId: (requested: unknown) => string;
 ```
 
 ### `warnIfSecondsLeading` (const)
--- a/examples/auth-playground/lunora/_generated/functions.ts
+++ b/examples/auth-playground/lunora/_generated/functions.ts
@@ -136,6 +136,27 @@ const callRegistered = async <R>(context: CallerCtx, functionPath: string, args:
         throw new LunoraError("FUNCTION_NOT_FOUND", `function not registered: ${functionPath}`);
     }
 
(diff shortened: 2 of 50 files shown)


prisis:

---

PR #579, opened 2026-09-02
fix: audit rounds 7-11

prisis:
Five rounds of bug-hunting and their fixes, merged into one change. **58 commits, 47 packages, 747 files.**

The chain that produced this had stopped being a chain — repeated squash-merges left no branch an ancestor of the next, so three of five reported conflicts against their own bases. The branches were re-merged oldest to newest so later work wins where the same file evolved twice, and 37 conflicts were resolved by hand. After regenerating the API snapshots the tree was byte-identical to the most complete chain branch, which is the check that the union is complete rather than merely conflict-free.

## What is in it

**Rounds 7-8** — payment refund accounting, adapter pagination, the bigint backfill and staged search indexes, x402 subscription refusal, client read-cache reuse, one bad subscription no longer taking the socket or shard with it, storage prefix matching on a segment boundary, rotated admin tokens no longer poisoning the queue, scaffolds no longer shipping a 403 home page.

**Rounds 9-10** — transactional composed mutations, identity claims actually forwarded by `getIdentity`, inbound mail dispatched as a trusted system call, backup snapshots written in the shape restore imports, search retrieval answering from the right vectors and columns, OTLP log redaction and batch caps, the Node platform host, the client replay path, and the eight non-JS SDKs.

**Round 11** — a branch audit of all of the above. Three were defects the stack had introduced itself: a conformance case that **deadlocked the real Cloudflare host** (gated timer continuations deliver in scheduling order, so a timer armed outside `blockConcurrencyWhile` before one armed inside it blocks the continuation the gate is waiting for); the **subscription cap cut to a quarter of what a socket holds**, derived against a 2,048-byte attachment budget when the measured workerd ceiling is 16,384; and **admin bulk operations metering per row**, so the ceiling reset every row and bounded nothing.

**Review rounds** — 110 CodeRabbit comments from the merged group PRs were triaged (79 had been overtaken by later rounds, 31 were still valid), then two review passes over those fixes found 13 more, then CodeRabbit reviewed the fixes and found 9 more. The most serious were introduced by the fix work itself:

- **A fail-open mail gate.** Keeping later authentication clauses removed the containment that first-match semantics gave against a quote-blind clause splitter — forged mail could reach the root shard's admin RPC with RLS disabled.
- **A truthy-verdict class across nine security gates**, found by grepping the shape rather than fixing the one reported: `adminGate` unlocked every admin route, the CORS predicate's `indexOf` returning `-1` allowed every origin, x402 auto-approved payments, a replayed turnstile token passed.
- **A unique-index guard applied to one twin only** — and the missing one is the default backend, where the failure is a shard that never opens again.
- **An appended event applied past an incomplete catch-up**, which advanced every materializer beyond the gap; since the watermark is a `min` over them, the skipped entries were never fetched again and the loss survived restart.

## Breaking changes

A squash collapses the individual declarations, so the full set is recorded here.

- **`client`** — `UsePaginatedQueryResult`, `UseInfiniteQueryResult`, `UseAgentResult`, `UseAgentChatResult` and `UsePresenceResult` gain `error`; `usePreloadedQuery` / `hydratePreloaded` take a second options argument. The registry's `pollIntervalMs` attach option is removed and `attach` now throws instead of degrading to polling.
- **`core`** — `PersistenceAdapter` requires a `replace(mutation)` method, and `PersistenceOperation` gains `"replace"`. A custom adapter must implement it atomically; the three built-in factories already do.
- **`do`** — `ShardDO.executeStream` takes the socket identity as a third parameter, `ShardDO.transactionHeadroom()` no longer returns `undefined`, and `MAX_SUBSCRIPTIONS_PER_SOCKET` is 8 and bounds subscriptions and shapes together.
- **`mail`** — `@lunora/mail/inbound` exports `authenticatesFrom`, and the playground's inbound gate moves from "DMARC or (SPF and DKIM) passed" to "some clause passed AND aligned with the `From` domain". An `authorize` / `adminGate` / `isAdmin` / `validate` / `onPaymentRequired` callback returning a truthy non-boolean now denies instead of allowing.
- **`observability`** — `ws-gate-open` is `warning`, not `error`, on a production worker; nothing in the audit emits `error` any more.
- **`payment`** — `PaymentAdapter.refundPayment` returns `RefundResult`; `WebhookAction` gains `refundId`.
- **`payment`** — `RefundResult` gains an optional `pending` flag; an adapter whose refunds settle asynchronously must set it or its refunds are recorded before the money moves.
- **`repo`** — `ConformanceHost` gains an optional `isolatesByDispatch`; a host that gates concurrent dispatches must set it or the isolation leg will deadlock against it. BREAKING CHANGE: `TOO_MANY_SUBSCRIPTIONS` and `SUBSCRIPTION_PERSIST_FAILED` message text changed, and the per-socket subscription cap moved from 8 to 32.
- **`repo`** — `InboundAuthentication` (`@lunora/mail/inbound`) replaces its six scalar fields (`dkim`/`dkimDomain`/`spf`/`spfDomain`/`dmarc`/`dmarcDomain`) with three lists of the new `InboundAuthResult` (`{ domain, result }`). RFC 8601 lets one `Authentication-Results` header report a method more than once and real mail does — an ESP-relayed message is DKIM-signed by both the relay and the author domain — and keeping only the first clause threw away the aligned pass whenever the MX listed the other one ahead of it. `@lunora/agent`'s inbound gate then bounced a message the receiving MX had fully authenticated. Consumers ask "does any reported clause pass and align?" instead of reading a single verdict; an empty list is "not reported".
- **`repo`** — `lunora view --remote` is removed. A deployed worker serves no studio, so the flag had no reachable target. The post-deploy summary's `studio: lunora view --remote` line is dropped for the same reason.
- **`repo`** — `serveJsonHandler`'s fifth parameter is a context object, not a `schemaDirectory` string; `scaffoldPolicyFile` / `wireRlsIntoProcedure` take the server module specifier; `resolveAdminToken` no longer reads `process.env`.
- **`repo`** — `serveStorageObject`'s fourth parameter is the `authorize` callback itself, not a one-field options object; `ServeStorageObjectOptions` is replaced by `StorageServeAuthorizer`. `@lunora/vite` no longer exports `DEV_WORKER_ENV_VALUE` / `DEV_WORKER_ENV_VAR`. `listLunoraSourceFiles` takes only a directory. The schedule-delay guard throws `INVALID_INPUT` where it threw `INTERNAL` or `BAD_REQUEST`.
- **`repo`** — `withDevWorkerEnv`, `createCommandProbe` and their exports are removed; `studioPlugin` takes the resolved plugin options.
- **`repo`** — a global table's physical column budget drops by one — the ceiling is now 97 declared fields on D1 rather than 98, reported by the same VALIDATION_ERROR.
- **`repo`** — a queued write that meets an envelope-less 4xx now rejects instead of being retried forever, and such a response rejects with the code its status maps to (FORBIDDEN, NOT_FOUND, PAYLOAD_TOO_LARGE, BAD_REQUEST) rather than INTERNAL.
- **`repo`** — the `lunora.prefix` module option is removed.
- **`repo`** — the coded error type gains a `transient` flag and `FlushReport` a `retry_after_ms` field; the durable record's `args` key now holds the wire form, so a store written by an earlier build hydrates as undecodable and is purged.
- **`repo`** — the third gate of `toolDefinitions`/`callTool` is `allowObservability`, not the presence of an admin token. A server that relied on the token alone must now pass `allowObservability: true` or set LUNORA_MCP_ALLOW_OBSERVABILITY.
- **`server`** — `runAt` rejects a non-finite instant with `INVALID_INPUT` (400); `assertScheduleDelay` takes a third `argument` parameter; `createCaller` refuses a mutation composed from a query context, as `ctx.runMutation` already did.
- **`server`** — `serveStorageObject(ctx, key, request)` now requires a fourth `{ authorize }` argument. `Storage` gains a readonly `bucketName`.
- **`testing`** — `AdvisorInput.tableSamples` and `constraintValidator` are removed; `ShardError` gains `code`; `serveStateless` takes `ServeStatelessOptions`.
- **`testing`** — `DeployCommandOptions` gains `emitBindings`.
- **`testing`** — `EventLogDOClient.getSince` returns a page; `getRange` is removed; paginated result types gain `error`.
- **`testing`** — `ShardDO.subscribe` can return `"paid"`; the emitted shard subclass overrides `isPaidFunction`.
- **`testing`** — `ValidatorIR.hasRefinement` is replaced by `refinements`/`refinementArgs`.
- **`testing`** — the recorded search-index profile now carries the field as well as the analyzer, so every deployed search index rebuilds once on the first migration after this ships, on both planes. The rebuild is in place and coverage stays latched, so it costs one backfill walk per index, not an outage. `BIGINT_KEY_NEGATIVE`/`BIGINT_KEY_NON_NEGATIVE` are no longer exported; `decodeBigintSqlKey` is.
- **`workflow`** — `@lunora/studio`'s `collectSkippedFkColumns` is now `collectUnresolvableFkColumns`, and `composeCreateIndex` is `composeIndexDeclaration(indexName, fields)` in `compose-index-declaration.ts`. The studio's statement splitter moved to `shared/sql-split-statements.ts`.
- **`workflow`** — `@lunora/workflow` exports `isDuplicateInstanceError`; `@lunora/agent` no longer re-exports it from `./channels`.
- **`workflow`** — `DoAuthWiring.resolveIdentity` resolves to a wider claim object.
- **`workflow`** — `PaymentStore.markEventProcessed` takes a third argument, the event type recorded on the claim row. A custom store implementation must accept it; the shipped database and in-memory stores are updated.
- **`workflow`** — `agent.asTool`'s default `maxPolls` is 600, not 120.
- **`workflow`** — a mutation composed from an action now commits or rolls back as one unit instead of autocommitting each write, and the jobs it schedules are enqueued only after that commit.
- **`workflow`** — snapshots written by the `backup` registry item change format. An existing `__table`-framed snapshot has to be converted before it can be restored.

## Verification

Run on the merged tree, not per branch: `build:packages`, `api:check` against all 54 snapshots, `lint:generated`, `lint:types` (122 projects), the full test sweep (65 projects / 175 tasks), `test:coverage` (56 projects / 152 tasks, zero threshold failures), prettier, eslint, `lint:package-json`, the secret scanner, the `LUNORA_WORKERD_TESTS=1` workerd suite, and a production build with `dist:check` over 1,378 shipped files.

## Two things to know before merging

**A search availability regression, kept deliberately.** On a deployment whose rows predate search-profile tracking, coverage is reset on the first backfill page, so every search on that index returns `SEARCH_INDEX_BUILDING` until the re-walk completes, where before it kept serving. Confirmed end to end with 1,200 rows. The trade is narrow but sound: serving rows analysed over an unknown column is silently wrong, while the refusal is a 503 whose message names `backfillSearch`, which closes it in one call. **This belongs in the upgrade notes.**

**CodSpeed reports one regressed benchmark.** The analyzer's accented-latin path goes 4.1µs to 4.5µs wall-clock, the measured price of two correctness fixes: the Cyrillic diacritic carve-out (`й` and `ё` decompose to separate letters and were being merged into the same index term) and the CJK presence test. Two ways to recover it were tried and both measured slower. The same run shows a +16.41% improvement on the d1 write-cost benchmark.

---


prisis:
@coderabbitai review

prisis:
@coderabbitai full review

PR #583, opened 2026-09-03
chore: open round 14 for review (do not merge)

prisis:
**Do not merge — this exists so CodeRabbit can review code it would otherwise skip.**

All of this is already on #579, which is 300+ files and therefore refused: `Review skipped: N files exceed the limit of 100`. A manual `@coderabbitai full review` accepts the command there and then refuses at review time on the same count, and there is no path-scoped review command. Slicing at a commit boundary is the only thing that works.

**Round 14** — 98 files. Packages that thirteen prior audit rounds never opened: hyperdrive, queue, seed, storage, notify, mcp, advisor, flags, browser, ai, cloudflare-access. 32 defects, the notable ones:

- **The paid MCP gate dropped `deps.waitUntil`** that both x402 siblings pass, so a settled payment's receipt sink was cancelled at isolate teardown — money moved on-chain, receipt row never landed.
- **The advisor's flag-polarity rule was inverted for every kill-switch flag.** `disableRls: true` scored clean while the safe setting was flagged with a remediation telling you to make it unsafe.
- **RAG `index()` hashed only the body**, so moving a document between tenants left the stale `orgId` on every vector.
- **`request.clone().json()` buffered without limit** and then entered the size check.
- **Six `@lunora/seed` divergences** where the `.unique()` path and the ordinary path disagreed — including timestamps seeded four decades apart for two spellings of the same column.
- **The MySQL parity suite had never run in CI**, skipping silently whenever mysqld was absent; it now executes 28 cases.

Four findings were rejected or narrowed with evidence rather than fixed.

---

## Summary by CodeRabbit

- **New Features**
  - MCP handlers support request-size limits, safer body parsing, client reuse, and expanded public types.
  - Push subscription registration reports replaced endpoints after VAPID key rotation.
  - Seed generation supports pinned timestamps and improved unique foreign-key values.
  - Storage worker examples now support CORS and safer content types.

- **Bug Fixes**
  - Improved security-lint detection, flag polarity checks, notification access scoping, email parsing, signed URLs, RAG filtering, browser cleanup, and replication ordering.
  - Added validation for oversized notification values and empty routing targets.
  - MySQL-backed test coverage is now required.

- **Documentation**
  - Clarified MCP opt-in, RAG scoring, browser keep-alive behavior, flag failures, and notification data exposure.

- **Breaking Changes**
  - Removed the seed `dateString` helper.
  - Updated notification and MCP handler interfaces.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -254,6 +254,91 @@
               "run": |
                   pnpm --filter "@lunora/${{ matrix.package }}" run test --project workerd
 
+    # `@lunora/hyperdrive`'s MySQL suites — the only gate proving the SQL the store
+    # core renders for a `.global()` MySQL backend actually executes. mysqld comes
+    # from `mysql-memory-server`, provisioned from npm like pglite and workerd
+    # rather than from a service container, so the job needs no `services:` block.
+    #
+    # The point of the job is that the suites cannot report green having run
+    # nothing. They skip when the binary is unobtainable, and no CI job had ever
+    # declared that mysqld was expected — so a suite that had never been observed
+    # to run reported success on every PR. `LUNORA_MYSQL_TESTS=1` turns that skip
+    # into a failure (`__tests__/_helpers/mysql-mem.ts`), and the assertion below
+    # closes the other half: a run that executes zero cases, or skips any, fails
+    # here instead of passing quietly.
+    "test-mysql":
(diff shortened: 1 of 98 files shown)


prisis:
Superseded — this work is on `alpha` via #586 (`6a09b746c`). This PR existed only as a review slice, since #586 was too large for a single CodeRabbit pass.

Findings it raised are recorded and being carried forward separately.

---

PR #585, opened 2026-09-03
chore: open round 16 for review (do not merge)

prisis:
**Do not merge — a review slice of #579.** See #583 for why this is split at all.

**Round 16** — 10 files. Four items earlier rounds fixed only halfway, plus one cross-branch repair:

- **The seed format fix was undone one line later.** A previous round made declared formats drive generation and refuse `.pattern()` at plan time, but the length axis still did a blind `value.slice(0, maxLength)` and a `padEnd`, so `v.string().email().max(20)` seeded an address its own validator rejects. Values that fit are now untouched, those that don't are rebuilt around the format, and a bound admitting no conforming value refuses at plan time naming the column — deterministic, because it is a property of the schema rather than of one generated value.
- **`MAX_BATCH_BYTES` was duplicated across 13 sites** with nothing reconciling them. The existing drift script was extended rather than a second mechanism added, and it compares **both operands**, so raising the worker's own body cap is caught too.
- **The registry deps sweep filtered to `.tsx?`** while the registry ships 114 Vue and Svelte components carrying 41 specifiers it never opened — a guard that passes by finding nothing, so a filter that stops matching a dialect reads as success.
- **The template smoke script recorded PASS with no compiler having read the source** on one path; it now fails closed.
- **A cross-branch repair**: a `@lunora/runtime` test regex-scrapes `toNdjson` out of `registry/backup/backup.ts`, and round 15 made that function span lines. The guard failed to load rather than reporting drift.

---

### DIFF
--- a/.github/file-filters.yml
+++ b/.github/file-filters.yml
@@ -234,13 +234,18 @@ templates:
     - ".github/workflows/test.yml"
     - ".github/file-filters.yml"
 
-# `MAX_BATCH_ENTRIES` in shared/batch-wire.ts vs the literal every non-JS port and
-# protocol/README.md restate. Separate from `sdks` because the reconciliation is a node
-# script, not a conformance suite: it must run on a change to any one of the ten copies
-# without paying for eight toolchains.
+# The two batch caps — `MAX_BATCH_ENTRIES` (shared/batch-wire.ts) and `MAX_BATCH_BYTES`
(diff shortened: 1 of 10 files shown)


prisis:
Superseded — this work is on `alpha` via #586 (`6a09b746c`). This PR existed only as a review slice, since #586 was too large for a single CodeRabbit pass.

Findings it raised are recorded and being carried forward separately.

---

PR #620, opened 2026-09-05
fix(protocol,sdks): pin the wire slots nine codecs answered differently

prisis:
Round-27 cross-port wave. Eight non-JS SDKs plus the TypeScript reference implement one wire codec; these are the slots where they disagreed, each one an unpinned leniency that nine implementations resolved several different ways.

## The reference moved twice

**base64 is validated for canonicity, not decodability.** Decode, re-encode, compare — `encode(decode(x)) != x` rejects. Spelled idiomatically but identically in all nine. Seven new fixture rejections (`unpadded`, `newline`, `whitespace`, `noncanonical-pad1/2`, `urlsafe`, `non-ascii`) each kill a different plausible shortcut.

This is a real wire-format change, so it was checked rather than assumed: no in-repo producer and no stored payload emits non-canonical base64. Every encoder is standard-padded (`btoa`, `base64.StdEncoding`, `b64encode`, `strict_encode64`, `Base64.getEncoder()`, `.base64EncodedString()`, `base64.encode`), and the persistence paths — SQLite JSON columns via `value-codec.ts`, CDC docs, aggregate keys — all route through those same encoders. `atob` already rejected the base64url alphabet, so `-_` was never accepted either.

**`url` hrefs must be absolute.** All eight ports stored the string verbatim and accepted `"not a url"`, where the reference's `new URL` throws — a frame that kills a JS peer's subscription and is waved through everywhere else. Ports get an RFC 3986 scheme floor rather than a WHATWG reimplementation; no stored relative URL is at risk, because a `URL` object serialises absolute and no JS producer could emit one.

The href fixtures were then found to be unpinned themselves: the three rejections were `"not a url"`, `""` and `"//example.com/x"`, all of which `href.includes(":")` satisfies. Added `":x"`, `"1http:x"` and `"é:x"` as rejections, plus `a+b-c.1:x` as an accepted counterweight so a port cannot satisfy them by narrowing to alphanumeric-only schemes.

## A round-trip assertion that could not see the bug

The conformance suites compared `canonical(roundTripped) == canonical(expected)`, and `canonical` normalises numbers the ECMAScript way — so a port writing `1.0` where the reference writes `1` compared **equal** and shipped the divergence. The comparison normalised away the very difference it existed to test.

Every suite now also asserts on the wire bytes its own transport would send, defined the same way in all eight. Measured, rather than assumed to help: the line is live in python, ruby, rust and dart, and inert-but-uniform in go, java and kotlin (their parsers map both sides to the same type). That measurement is recorded in `sdks/README.md` rather than left implied.

## Other divergences

- `-0` epoch handling: six ports patched, three correctly left alone because their type systems already produce the right answer. The fixture is written as `-0.5`/`-0.0` so a JSON parser cannot narrow away the sign before the codec sees it.
- Java's `Map.entry` NPE on a null value, and Python's bare `ValueError` where `binascii.Error` was expected — both error-type bugs, not just crashes, driven by fixtures so all nine are held to them.
- `encodeWire` could produce a frame `decodeWire` refuses: the error `name`/`message` slots were written verbatim while the new decode guard rejects a non-string in either. A server-side error with a non-string message made the client's decoder throw instead of surfacing the error — and a decoder throw on a subscription frame kills the subscription. Coerced symmetrically on encode, with tests in the three languages whose labels can hold a non-string.
- Rust's decoder carried an ASCII-whitespace strip that the canonicity compare made unreachable, under a comment claiming the opposite. Removed. The padding loop beside it is **not** dead — deleting it fails `conformance_manifest_is_covered` immediately — and now says so.

## A blind range that stays blind, honestly

Dart's `_encodeDouble` claimed a 2^53 bound "because past it `jsonEncode` spells a double in exponent form". Measured: exponent form starts at 1e21, and the whole `(2^53, 1e21)` range diverges by a trailing `.0`. Dart's bound cannot be raised — `(1e20).toInt()` saturates at `9223372036854775807` — so the comment now states the real bound and the residual gap instead of claiming it closed. A `1e20` fixture case was added and earns its place: setting the bound to the claimed `1e21` makes it fail, and nothing else in the fixture fires.

## Verification

`sdks/run-all.sh` 8/8 (python 100, go 252, ruby 78, rust 9, swift 11, java 489, kotlin 494, dart 82), `sdks/lint-all.sh` 8/8, `generated-check.sh` pass. Full repo gates green: `build:packages`, `lint:types`, `test`, `lint:prettier`, `lint:eslint`, `api:check`, `lint:package-json`.

Fixture enforcement was proven rather than inferred: poisoning one entry — moving a rejection into `cases[]` — fails all eight suites, and the corrected file passes all eight. Note that `run-all.sh` reports each runner's own test count, not fixture entries, so a flat count is not evidence a suite ignored a change. The wire-text assertion was proven the same way, by breaking each port's number formatting and confirming which suites go red.

## Reviewed before opening

A `/thermos` branch audit and code-quality pass ran against this branch first; their findings are fixed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error values now consistently encode their names and messages as strings.
  * Invalid error labels are rejected during decoding instead of being silently converted.
  * Non-canonical or malformed base64 payloads are now rejected during decoding.


prisis:
The rebase onto current `alpha` fixed the comparison base — this run compares `fix/r27-sdks (2c039e5)` against the real `alpha (7b26308)` rather than the 43-commit-stale substitute the earlier report fell back to. The remaining regression is environmental, and CodSpeed says so itself at the top of the report: *"Different runtime environments detected — Some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results."*

**The flagged benchmark cannot be reached by this diff.** Outside `sdks/` and `protocol/`, the branch changes exactly two files: `packages/client/__tests__/wire-codec.test.ts` and `shared/wire-codec.ts`. The codec is genuinely reachable from `trigger-overhead.bench.ts` (via `createShardCtxDb` → `ctx-db.ts`/`do-sql.ts`), so this was worth checking rather than waving off. Stripping comments, the entire code change is three things:

- `String(error.name)` / `String(error.message)` — the **encode** path, errors only
- a `typeof … !== "string"` guard on the same two slots — the **decode** path, errors only
- the base64 canonicity re-encode-and-compare — the **decode** path, `bytes` only

A `1 after-insert no-op trigger` write touches none of them. It encodes a plain-JSON document, and `encodeWire` is byte-identical for plain JSON — that identity is what the 68 accepted fixture cases pin.

The corroboration is in the same table: three unrelated benchmarks simultaneously "improved" 11–18%, including `v.string()` at +18.32% — the same benchmark that appeared as a *regression* on the pre-rebase run of a different branch, whose module closure is two files and touches neither. A diff that cannot reach four benchmarks did not move four benchmarks in two directions.

Nothing to fix here. Flagging one thing for whoever owns the CI config: `packages/do/__bench__/trigger-overhead.bench.ts` seems to sit close enough to the noise floor that a ~10% swing is within run-to-run variance on this runner, which makes it a recurring false positive rather than a signal.

PR #621, opened 2026-09-05
fix(errors,values,server): close the gates that were failing open

prisis:
Round-28 foundation wave. The through-line is checks that passed by not looking.

## The error catalog failed open, and its gate could not see why

`isInternalCode` is `getCatalogEntry(code)?.internal === true`, so an **unregistered code is treated as client-safe**. The registration gate that should have caught that matched only the literal `new LunoraError(`, making every `LunoraError` subclass mint invisible to it — including all six `@lunora/payment` codes, one of which carries "webhook secret not configured".

The scanner now sees indirect mints (subclass, `raise(`, `super(`, object literals), and the six codes are catalogued. `CONFIG_INVALID` is `internal`; `PROVIDER_ERROR` deliberately is not — its three mint sites are fixed capability text or echo back the caller's own id.

The scanner also got a self-check, for a reason worth stating: a repo-wide scan only goes red while a violation happens to exist, so once the catalog is complete a narrowing of the pattern back to the literal name lands **green**, and every future subclass mint is client-safe again.

Two follow-ups from review:

- `CONFIG_INVALID: { internal: true }` was over-redacting. `check()`'s argument-shape guard minted the same code for "check() requires a featureId or priceId" — the caller's own mistake — so marking the code internal turned actionable guidance into a generic 500. Those two sites now mint `VALIDATION_ERROR` (already in the payment taxonomy at 400, already catalogued as caller-safe). The `entitlements`-not-configured sites stay internal; they do name server configuration.
- The webhook-rejection entry claimed "the messages are fixed and name no internal state". Five of six mints are fixed strings; `providers/stripe.ts` forwards the Stripe SDK's verification message. The verdict is still right — those texts describe the poster's request — but the justification was not.

## `sql-readonly` refused a valid read

`SELECT REPLACE(name, 'a', 'b')` was rejected because `replace` was on the write-verb list. Now a lookahead — `\breplace\b(?!\s*\()` — which distinguishes the scalar function from `REPLACE INTO`. It is a lookahead rather than `replace\s+into` because a block comment between the two words is whitespace to SQLite's tokenizer but not to `\s`; the residual false refusal is named in the comment. Confirmed unwalkable: `REPLACE/**/INTO` still matches, and a statement mixing both still matches on the write.

## `toJsonSchema` disagreed with the runtime parser

Four counts, each fixed with a paired assertion proving the old expectation was lying — `expect(v.bigint().safeParse(5).ok).toBe(false)` sitting beside the schema assertion. `acceptsAbsent` replaces `kind !== "optional"` with a recursive "does the parser accept this absent" predicate, deleting a whole class of spec-vs-parser disagreement.

One honest limit, corrected after review: `v.bigint()`'s schema names the payload's form, not a body that validates end to end. Every RPC arg path runs `decodeWire` first, so a bigint's actual carrier is the tagged array `["$lunora.wire$","bigint","5"]` and a bare `"5"` fails the parser exactly as a bare `5` did. The emitted shape is unchanged and consistent with how `bytes` and `date` are modelled; only the comment claiming parser agreement is gone.

## Smaller

- The playground's `inbound:onEmail` is now `internalMutation`. The real mail path is unaffected — it POSTs with `x-lunora-system: 1` and dispatches by string path, so there is no typed reference and no client reference.
- A fake e2e DO-reset was deleted rather than implemented: it asserted nothing and existed to look like coverage.
- The docs deny-table gained a self-check, for the same reason the catalog scanner did. Fixing it also fixed a false positive it had: it rejected the **correct** non-destructuring form `const m = useMutation(fn)` then `await m.mutate(args)`. The two forms differ in whether the bound name is later called, which spans two lines, so the scan is now per-fence with a backreference. A CI gate that blocks valid docs is worse than the misuse it catches.

## Verification

Full repo gates: `build:packages`, `lint:types`, `test` (175 tasks), `lint:prettier`, `lint:eslint`, `api:check` (54 snapshots), `lint:package-json` — all green. Package suites: errors 669, values 187, payment 252.

Every doc-gate change was probed against all four real snippet forms; the deny-table's patterns were broken in both directions to confirm the self-check reports a narrowed pattern and an over-matching one.

## Reviewed before opening

A `/thermos` branch audit and code-quality pass ran against this branch first; their findings are fixed here. Three of the wave's original claims were rejected during the fix pass rather than implemented — `.nullable()` already declares the predicate param as `T|null` (verified with a TS18047 probe), and `ctx.db.from(...)` appears nowhere in the codebase.

Known residual, left deliberately: the catalog scanner's patterns are broader than their stated intent (`new TypeError("SOME_CONST")`, any `raise(`). That is the fail-safe direction — it over-collects and an allowlist handles the rest — but it will churn.

### DIFF
--- a/api-snapshots/errors.api.md
+++ b/api-snapshots/errors.api.md
@@ -240,6 +240,31 @@ const ERROR_CATALOG: {
         readonly status: 502;
         readonly title: "Cloudflare Workflows REST API error";
     };
+    readonly CONFIG_INVALID: {
+        readonly internal: true;
+        readonly status: 500;
+        readonly title: "Payment configuration invalid";
+    };
+    readonly CURRENCY_MISMATCH: {
+        readonly status: 400;
+        readonly title: "Currency mismatch";
+    };
+    readonly PROVIDER_ERROR: {
+        readonly status: 502;
+        readonly title: "Payment provider error";
+    };
+    readonly WEBHOOK_EVENT_ID_MISSING: {
+        readonly status: 400;
+        readonly title: "Webhook event id missing";
+    };
+    readonly WEBHOOK_SIGNATURE_INVALID: {
+        readonly status: 400;
+        readonly title: "Webhook signature invalid";
+    };
+    readonly WEBHOOK_TIMESTAMP_INVALID: {
+        readonly status: 400;
+        readonly title: "Webhook timestamp outside tolerance";
+    };
     readonly ADMIN_FORBIDDEN: {
         readonly status: 403;
         readonly title: "Admin access forbidden";
--- a/api-snapshots/values.api.md
+++ b/api-snapshots/values.api.md
@@ -203,6 +203,7 @@ interface SchemaNodeReader<TNode> {
     constraints: (node: TNode) => JsonSchema | undefined;
     inner: (node: TNode) => TNode | undefined;
     isNullable: (node: TNode) => boolean;
+    keyChild: (node: TNode) => TNode | undefined;
     kind: (node: TNode) => ValidatorKind;
     literalSchema: (node: TNode) => JsonSchema;
     members: (node: TNode) => ReadonlyArray<TNode>;
--- a/apps/docs/scripts/check-doc-imports.mjs
+++ b/apps/docs/scripts/check-doc-imports.mjs
@@ -1,13 +1,26 @@
 /**
- * Doc snippet import gate.
+ * Doc snippet API gate.
  *
  * Twoslash can't type-check the doc snippets: apps/docs doesn't depend on the
  * @lunora/* packages and the app-local `@/lunora/_generated/*` modules don't
- * exist here, so nothing resolves. This lighter check catches the bug class
- * that actually shipped — importing a symbol from a package that doesn't export
- * it (e.g. `import { query } from "@lunora/server"`, where the builders live in
- * the generated server). It parses every ```ts/```tsx fence across the docs and
- * verifies each named import against the target module's real exports.
+ * exist here, so nothing resolves. This lighter check catches the two bug
+ * classes that actually shipped:
+ *
+ *   1. importing a symbol from a package that doesn't export it (e.g.
+ *      `import { query } from "@lunora/server"`, where the builders live in the
+ *      generated server) — checked against the target module's real exports;
+ *   2. a symbol that IS exported, used with a shape it does not have — see
+ *      {@link NONEXISTENT_API}. Names alone can't see this class, and it is the
+ *      worse of the two: the flagship tutorial taught `useMutation` as a
+ *      callable for as long as the hook has returned an object, so a reader
+ *      following it got `TypeError: send is not a function`.
+ *
+ * It parses every ```ts/```tsx fence across the docs and applies both.
+ *
+ * A deny-table, not a type checker: it holds only shapes that were WRONG in
+ * shipped docs, each with the correct form in its message. Type-checking the
+ * samples for real needs the packages installed here plus synthesized
+ * `_generated` modules — worth doing, but not what this file is.
  *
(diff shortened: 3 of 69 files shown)


prisis:
Both CodSpeed regressions are artifacts of the substituted comparison base, not this branch. Evidence, since "pre-existing" is not something to assert without it:

**CodSpeed says so itself**, in its own footnote: *"No successful run was found on `alpha` (9a15365) during the generation of this report, so b00f68d was used instead as the comparison base. There might be some changes unrelated to this pull request in this report."* `alpha` has since moved 43 commits past `b00f68d`, so the diff being measured is mostly not mine.

**`v.string()` (-14.7%) cannot come from this branch.** The benchmark imports only `../src/v`. Computing that module closure gives exactly two files — `src/v.ts` and `src/errors.ts` — and this branch changes neither. It changes `src/json-schema-core.ts` and `src/to-json-schema.ts`, which `v.ts` deliberately does not import (there is a comment in `v.ts` saying it is kept structurally identical and local precisely so it never does). Measured locally on the branch, `v.string()` runs at 48.5M hz, indistinguishable from `v.number()` (48.1M), `v.boolean()` (48.4M) and `v.id()` (48.4M) — all six primitives sit within noise of each other, which is not what a regression in one of them looks like.

**`broadcastDelta — low fanout` (-73%) is not on a path this branch touches.** `shard-do.ts` does import `sql-assistant.ts`, which this branch modifies, so the link is not absurd — but the change is entirely inside the statement-classification branch that runs only when the model has generated SQL (it adds a `console.warn` before an existing `return undefined`). `broadcastDelta` never calls `generateSql`/`generateFilter`/`generateChart`, and no module-level work was added.

A 3.7× regression in socket fan-out from adding a warning to an admin-gated SQL path is not a plausible mechanism; a 43-commit base delta is.

Happy to rebase onto current `alpha` if a real baseline is wanted, but I have deliberately not: `.git/info/attributes` maps both `pnpm-lock.yaml` and `api-snapshots/*.api.md` to a `keepours` merge driver, which resolves those silently rather than conflicting, and `alpha` has touched both. CI builds `refs/pull/N/merge`, so the merged state is already what the green checks reflect.

prisis:
The rebase onto current `alpha` fixed the comparison base, so these numbers are now measured against the real `alpha (7b26308)`. The remaining regressions are runner variance, and the cross-branch evidence is conclusive rather than a judgement call.

**`normalizeLogFields :: flat 3 primitives` is reported as a regression on three unrelated branches at once**, all from the identical `55.6 µs` base: #618 (telemetry — observability/agent/runtime), #619 (codegen + scripts) and #621 (errors/values/payment). Those three share no code. **None of them changes `packages/do/__bench__/normalize-log-fields.bench.ts` or anything it measures** — `git diff --name-only origin/alpha...<branch> | grep -c normalize-log-fields` is `0` for all three.

The sharpest case is #619: it changes **zero** files under `packages/do`, and is credited with −72.84% on `broadcast-delta` and −52.72% on `normalize-log-fields`, both `packages/do` benchmarks. A diff that touches no file in a package cannot move two benchmarks in it by 50–70%.

Two more corroborations from the run history:

- `broadcast-delta — low fanout` at `114 µs → 42x µs` was attributed to #621 on the pre-rebase run and to #619 on this one — the same benchmark, the same magnitude, two different branches.
- `v.string()` was a **regression** on one pre-rebase run and an **+18.32% improvement** on #620 now. Its module closure is two files (`src/v.ts`, `src/errors.ts`), which no branch here touches.

CodSpeed flags this itself at the top of each report: *"Different runtime environments detected — Some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results."*

Nothing to fix on these branches. Worth someone owning separately: `normalize-log-fields` and `broadcast-delta` appear to sit close enough to the noise floor on this runner that they produce recurring false positives, which is expensive because each one has to be disproved by hand.

PR #631, opened 2026-09-06
fix(scheduler,workflow): bracket the scheduler args with the wire codec

prisis:
Round-31. `ctx.scheduler.runAfter/runAt` args never went through the wire codec at the producer, while one of two consumers decoded them — so a `bigint` arg threw before the job was ever recorded, and a `Date` silently became a string.

## The asymmetry

Producers sent raw: `create-scheduler.ts`, `create-workpool.ts`, `create-worker.ts`'s `/schedule` post, and `mail/inbound/shard.ts`. Consumers disagreed: a function target routes through the shard, which decodes at `shard-do.ts:5925`; a workflow target goes to `create({ params })` and `packages/workflow` contained **no** `decodeWire` at all.

That asymmetry is why this was left open in round 30 rather than "just encoded at the producer": doing so alone would have fixed the shard path and pushed raw tagged arrays into every scheduled workflow's `params`.

## The fix, and where it ended up

Both sides are now bracketed, and after review the codec moved to the **chokepoints** rather than sitting at call sites: `do-client.ts`'s `callDO` encodes and `requestDO` decodes, so `/schedule`, `/cancel`, `/list`, `/get`, `/dead`, `/dead/retry` and every future route are correct by construction. Same for `buildHttpScheduler`'s `post`/`call`. The codec is identity for plain JSON, so `functionPath`, `scheduledFor`, `cursor` and friends are byte-identical.

That was not cosmetic. The first version encoded at the producer and left `list()`, `get()` and `dead()` un-decoded **twenty-six lines below in the same file** — improving dispatch while regressing read-back, on the documented public `Scheduler` surface. Call sites went from 9 scattered to 8 with 2 chokepoints; undecoded read routes from 5 to 0. This defect class has now been found four times in this campaign, which is the argument for a chokepoint over a convention.

The workflow side decodes where the handler reads `params`, not at dispatch: Cloudflare JSON-serialises `params` into durable storage, so the wire form is exactly what survives that hop, whereas a real `bigint` fails instance creation and a `Date` degrades to a string.

## Three defects found while fixing it

- **`ctx.scheduler.get()` returned the wrong shape** in the runtime facade — the DO's `{ record }` envelope rather than the record, so `job.scheduledFor` was `undefined`, and a truthy `{}` on a miss instead of `null`, so `if (job)` reported a cancelled job as still pending.
- **A run-context decode failure was retryable.** Workflow params are already durable, so every retry decodes identical bytes to the identical failure and burns the whole budget. Now non-retryable.
- **`decodeWire` is not idempotent.** Once the codec moved into `do-client`, the existing `system-reader` decode became a *second* one — and a second decode flattens a `Date` to `{}`, because a `Date` has no own enumerable keys (a `bigint` survives). Verified: `once: { dueAt: 2026-06-01T…, n: 42n }` → `twice: { dueAt: {}, n: 42n }`. That decode is removed, with a comment naming the trap.

## Verification

Every regression test was run against the unfixed code first, and the new ones drive the **consumers** rather than restating `decodeWire(what the producer sent)` — which is precisely why the read-back hole passed every gate the first time round. Eight before/after pairs, including `expected [ '$lunora.wire$', 'date', … ] to be an instance of Date` on `list`/`get`/`dead`.

All seven repo gates green, `api:check` with no drift.

`packages/mail` is bracketed at `postShardRpc`, its single wire owner. One behaviour break noted in its commit body: a caller-supplied `resolveArgs` override returning a class instance previously delivered a lossy `{}` and now throws, because the codec refuses non-plain objects.

## Still open, deliberately

`createDispatchRunner` ships `args` un-encoded to the same decoding shard, so `ctx.run(fn, { n: 1n })` is broken in `@lunora/queue`, `@lunora/workflow` and `@lunora/agent`. Fixing it in the shared runner is **mutually exclusive** with the queue-workpool bracket on a sibling branch — it would double-encode — so it belongs at each package's own `run` wrapper and is filed rather than smuggled in here.

### DIFF
--- a/packages/client/__tests__/lunora-client.test.ts
+++ b/packages/client/__tests__/lunora-client.test.ts
@@ -2810,6 +2810,25 @@ describe("lunoraClient", () => {
             await expect(client.listScheduledJobs()).resolves.toEqual([]);
         });
 
+        it("listScheduledJobs raises WIRE_DECODE_FAILED on args the codec refuses", async () => {
+            expect.assertions(2);
+
+            // `decodeWire` throws a bare `TypeError` on a malformed tag and a
+            // `RangeError` past its depth bound, and can decode to a non-object.
+            // Both used to escape unmapped — and inside `subscribeScheduledJobs`
+            // they landed in a `catch` labelled "a non-JSON frame", which silently
+            // froze the operator's job list.
+            const malformed = [{ args: ["$lunora.wire$", "bigint", "not-a-number"], enqueuedAt: 1, id: "j1", scheduledFor: 2 }];
+            const client = new LunoraClient({
+                fetch: async () => jsonResponse({ records: malformed }),
+                url: "https://app.example",
+                WebSocket: createMockWebSocket(),
+            });
+
+            await expect(client.listScheduledJobs()).rejects.toMatchObject({ code: "WIRE_DECODE_FAILED" });
+            await expect(client.listScheduledJobs()).rejects.toThrow(/j1/u);
+        });
+
         it("cancelScheduledJob POSTs the id and normalises the result", async () => {
             expect.assertions(4);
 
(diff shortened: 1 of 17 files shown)


prisis:
Closing — 113 commits behind, and split between work that already shipped and one design decision that went the other way. The live remainder is in #644.

**Commit 1 (`put job args through the wire codec`) is fully on `alpha`, in a better form.** Every conflict hunk on rebase showed `alpha` already doing the same thing with `encodeArgsOrThrow` — which labels the call and names the function path — rather than this branch's bare `encodeWire(...) as Record<string, unknown>`. Same for `decodeScheduleRecords` (shipped as `decodeRecordArgs`) and the `run-context.ts` decode.

**The `do-client.ts` change is a competing design, not a missing fix.** It moves encode ownership to the DO transport (`callDO` encodes, `requestDO` decodes) and correspondingly *shrinks* `create-scheduler.ts` and `create-workpool.ts`. `alpha` shipped the other design — each producer encodes (`create-scheduler.ts:73`, `create-workpool.ts:59`) and then calls `callDO`. Applying the transport encode on top of that would double-encode, which is exactly the defect #643 just repaired on the queue path.

**`system-reader.ts` landed verbatim** (`args: (record["args"] as Record<string, unknown> | undefined) ?? {}`).

**What was genuinely still live, and is now in #644:**
- the scheduled-jobs socket discarding `decodeWire` failures under an empty `catch` labelled "a non-JSON frame", freezing the operator's job list in total silence
- a workflow's params decode raising a bare retryable `TypeError`/`RangeError` on a payload that is permanently undecodable, burning the whole retry budget
- `fake-scheduler` rejecting a `bigint` arg that production now carries, under a comment whose premise stopped being true when the producers began encoding

Each of those was re-run against the unfixed code and observed to fail before being fixed.

PR #635, opened 2026-09-06
security(runtime): let a declared proxy header restore per-IP limiting

prisis:
> **Stacked on #630** (`fix/r30-client-ip-trust`) — based on that branch, not `alpha`, so this diff is only the opt-in. Merge #630 first.

## Why

#630 stops the runtime trusting `cf-connecting-ip` off the edge, which is correct — off Cloudflare it is a client-typed header, and trusting it removes the limit entirely. Its consequence, raised in review and left documented there: off the edge the REST limiter's default key collapses to one shared bucket and `ctx.ip` is `undefined`.

For a bare Node host that is no change; the header was already absent. For a `@lunora/platform-node` process **behind** Cloudflare — where the edge does stamp it and the origin only accepts CF traffic — it trades "an attacker escapes their own limit" for "an attacker can lock out every honest caller". A different failure mode, not a smaller one.

#630's answer is "pass an explicit `key`". That covers the REST limiter and has **no equivalent for `ctx.ip`**, which has no callback — so such a deployment could not recover `ctx.ip` at all. That gap is what earns this option beyond a knob.

## The shape

`trustedClientIpHeader?: string`, on `WorkerOptions` (governing `ctx.ip`) and on `createRestRateLimit` (governing the limiter's key). On the edge, `cf-connecting-ip` still wins regardless. Off the edge with nothing declared, the fail-closed default from #630 is unchanged — the opt-in is the only thing that alters it.

It mirrors `@lunora/auth`, which already gates on `onCloudflareEdge()` and re-admits a header only once an operator declares `trustedProxies`. **It takes a header name rather than auth's proxy-address list, and that is forced rather than stylistic**: `fetch` hands the runtime a `Request` and nothing else. Grepping `remoteAddress|remoteAddr|socket.remote|peerAddress` across `runtime`, `platform` and `platform-node` returns **zero** hits — there is no peer address to match an address list against. better-auth can accept `trustedProxies` because it walks an XFF chain past declared hops; this layer has no chain walker.

That is also why a comma-bearing value is refused: that is the appended-chain shape whose leftmost entry is client-written, and reading it as one address would hand back exactly the attacker-chosen string this module exists to refuse.

## Verification

Four cases added, run against the unfixed tree first. Two failed for the right reason — the limiter keying both callers into `"no-trusted-ip"`, and `expected null to be '203.0.113.7'` for `ctx.ip`. **The other two pass vacuously against unfixed code** and are flagged as such in the report rather than counted as regressions: with no opt-in support the answer is already "resolve nothing" and "cf-connecting-ip wins". They are guards against the fix over-reaching, which matters here — a fix that trusted the header everywhere would re-open the original hole.

All seven repo gates green; `api:check` drifted exactly by the two new fields and the snapshot is committed.

## One rough edge, kept rather than papered over

`createRestRateLimit` builds its closure before `createWorker` is called, so the worker cannot inject its own value without changing the `RestRateLimit` contract. Both docblocks and the docs example say to set both, or the two disagree about who a request came from. Unifying them means restructuring that seam, which is worth a reviewer's opinion rather than a silent decision.

Also fixed here: `concepts/rest-api.mdx` still said the default "keys on the caller's IP (`CF-Connecting-IP`)" unconditionally — already wrong as of #630, which missed it.

### DIFF
--- a/api-snapshots/runtime.api.md
+++ b/api-snapshots/runtime.api.md
@@ -2257,6 +2257,7 @@ interface WorkerOptions {
     storageSignedUrl?: StorageSignedUrlFunction;
     storageUpload?: StorageUploadFunction;
     syncGlobals?: GlobalCdcSyncFunction;
+    trustedClientIpHeader?: string;
     trustInboundTraceContext?: TrustInboundTraceContext;
     vectorIntrospector?: VectorIntrospector;
     voiceAgents?: Record<string, ShardNamespaceLike>;
@@ -2395,6 +2396,7 @@ const createQueryCoordinator: (options: QueryCoordinatorOptions) => QueryCoordin
 const createRestRateLimit: (limiter: RateLimiterLike, options: {
     key?: (request: Request, functionPath: string) => string | undefined;
     name: string;
+    trustedClientIpHeader?: string;
 }) => RestRateLimit;
 ```
 
--- a/apps/docs/src/content/docs/concepts/rest-api.mdx
+++ b/apps/docs/src/content/docs/concepts/rest-api.mdx
@@ -135,9 +135,16 @@ export default createWorker({
 ```
 
 It runs **before** dispatch, so a limited request costs you nothing downstream. By
-default it keys on the caller's IP (`CF-Connecting-IP`); pass `key` to key on
-something else, such as an API key header, or the `functionPath` to limit per
-endpoint:
+default it keys on the caller's IP from `CF-Connecting-IP`, but only while running
+on Cloudflare, where the edge stamps that header over anything the client sent.
+On any other target nothing overwrites it, so it is a header the caller typed and
+trusting it would hand an attacker a fresh bucket per request — those deployments
+resolve no IP and pool into one shared `no-trusted-ip` bucket instead. An origin
+fronted by a proxy that does stamp a client address declares it with
+`trustedClientIpHeader` to get per-IP buckets back; see
+[`@lunora/runtime`](/docs/packages/runtime) for what that assertion commits you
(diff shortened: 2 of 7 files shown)


prisis:
Rebased onto current `alpha` (was 113 commits behind), and collapsed to what is actually left.

Nine of this branch's ten commits had already landed — the `shared/on-cloudflare-edge.ts` predicate, the default-deny `trustedClientIp` base, the auth where-clause partition, the expo notes and the test cleanups are all on `alpha` now. Rebasing them produced conflicts against their own merged descendants, so I reset to `alpha` and cherry-picked the one commit that is genuinely new: `security(runtime): let a fronted origin declare its client-ip header`.

The remaining feature is unchanged — `trustedClientIpHeader` as the operator's opt-out for an origin behind a proxy that stamps a client address, ignored on the edge where `cf-connecting-ip` still wins, comma-bearing values refused.

Two notes:
- My first attempt copied the branch tip's files wholesale and broke `@lunora/server`'s build (`"SpanContextIds" is not exported by "src/types.d.ts"`) — `alpha`'s `types.ts` had gained an export the old copy did not have. Redone as a cherry-pick so only the feature's own hunks apply.
- `api-snapshots/runtime.api.md` auto-merged during the cherry-pick, which this repo's `keepours` merge driver does silently. Regenerated it from a fresh build rather than trusting that; it produced a further change, so the merged version was indeed not what a rebuild yields.

Gates green: `build`, `test`, `lint:types`, `lint:eslint`, `lint:prettier`, `api:check`, `lint:package-json`, `check-generated-files`.

PR #636, opened 2026-09-06
feat(codegen): discover the crons configured on createWorker

prisis:
> **Stacked on #629** (`fix/r30-schedule-provisioning`) — based on that branch, not `alpha`, so this diff is only the discovery. Merge #629 first.

## Why

#629 stops `lunora deploy` deleting hand-written `triggers.crons` entries, by recording which crons codegen generated in `package.json` → `lunora.crons`. That record exists because two cron surfaces are invisible to codegen: `createWorker({ backupCron })` and `createWorker({ crons })`, both documented as needing a hand-written wrangler entry.

This makes the common case visible. A `backupCron: "0 3 * * *"` literal in the worker entry is now a cron codegen knows it owns — so it is kept for the right reason rather than merely preserved as "not ours", and it is **cleared correctly when removed**, which the ownership record alone cannot do for it.

## The record stays — the proposal to delete it does not hold

The review that suggested this discovery argued it would make the generated set complete and let the ownership record be deleted. It cannot.

`.extend(fn)` is the generated escape hatch: `(env, derived) => Partial<WorkerOptions>`, merged straight into the `createWorker` options. Both `backupCron` and `crons` are members of `WorkerOptions`, and the emitted `AppBuilder` applies every `extendFn` to the built options at runtime. So `.extend((env) => ({ backupCron: env.NIGHTLY_CRON }))` is legitimate and no AST scan can resolve it — discovery is **structurally** incomplete, not incidentally so. The record is the safety net for exactly that residue; removing it would reintroduce #629's deletion bug for dynamically-configured apps, with a smaller blast radius and no signal.

## What it reads, and what it cannot

`packages/codegen/src/discover/worker-entry-crons.ts` scans the existing worker-entry file set — the same `listSecurityScanFiles` five security lints already use — for `createWorker({…})` option literals and for the literal an `.extend(fn)` callback returns.

Reachable: `backupCron` as a string or untagged template literal, on either surface, with a concise body or a single-`return` block; `crons` keys as string literals, including the shorthand-method form. Deduped, then unioned into the `cronTriggers` both the reconciler and the CLI's Cron-Trigger-limit warning consume.

Out of reach, each pinned by a test asserting **nothing** is found: a variable or member expression, a substituted template, a computed or identifier `crons` key, a spread map, an opaque options object, a multi-statement `.extend` body, and a `createWorker` call outside the entry roots.

`objectLiteralFromCallbackBody` moved into `discover/ast.ts` rather than being duplicated — both readers of `.extend(fn)` must agree on which bodies are statically readable.

## Verification

Against unfixed code the two positive tests fail for the right reason (`expected [] to strictly equal [ '0 4 * * *' ]`, and the missing `"0 3 * * *"` in the trigger set); the computed-`backupCron` test passes both before and after, which is correct — it must never be discovered.

The interaction is pinned where it matters: `reconcile-crons.test.ts`'s existing preserve test was retitled from "a hand-written trigger the project's crons.ts never declared" to "a trigger codegen could not derive — a computed backupCron", because its old rationale is now false for half the cases while its assertions still cover the right half.

Fixtures are codegen unit plus a `runCodegen` integration test that writes a real `src/server/index.ts` and runs the whole wiring — rather than putting a `backupCron` into a demo app with no use for one. No example sets these options, so `check-generated-files.mjs` reports no regeneration.

All seven repo gates green.

## Considered and skipped

Folding `entryCronTriggers` into the platform-gate signal — a `backupCron` on a host whose matrix rates `cronTriggers` unsupported is inert, so it is arguably honest, but it changes gate output for apps that pass silently today and no `ctx.*` surface is being added, so the platform-parity rule does not compel it.

## Summary by CodeRabbit

* **New Features**
  * Automatically includes literal worker backup and scheduled cron configurations in generated deployment triggers.
  * Supports cron schedules configured directly on worker definitions and extensions, with duplicate schedules removed.

* **Documentation**
  * Clarified that dynamically computed or spread cron configurations must be added to deployment triggers manually.

### DIFF
--- a/packages/codegen/__tests__/discover/worker-entry-crons.test.ts
+++ b/packages/codegen/__tests__/discover/worker-entry-crons.test.ts
@@ -1,0 +1,149 @@
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
+
+import { Project } from "ts-morph";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import discoverWorkerEntryCrons from "../../src/discover/worker-entry-crons";
+
+let workdir: string;
+let project: Project;
+
+/** Write a project-relative source file — the worker entry lives outside `lunora/`. */
+const writeAt = (relative: string, source: string): string => {
+    const path = join(workdir, relative);
+
+    mkdirSync(dirname(path), { recursive: true });
+    writeFileSync(path, source, "utf8");
+
+    return path;
+};
+
+const discover = (): string[] => discoverWorkerEntryCrons(project, join(workdir, "lunora"));
+
+describe("discoverWorkerEntryCrons", () => {
+    beforeEach(() => {
+        workdir = mkdtempSync(join(tmpdir(), "lunora-entry-crons-"));
+        mkdirSync(join(workdir, "lunora"), { recursive: true });
+        project = new Project({ skipAddingFilesFromTsConfig: true, useInMemoryFileSystem: false });
+    });
+
+    afterEach(() => {
+        rmSync(workdir, { force: true, recursive: true });
+    });
+
+    it("reads a string-literal backupCron from the worker entry", () => {
+        expect.assertions(1);
+
+        writeAt("src/server/index.ts", `export default createWorker({ backupCron: "0 3 * * *", backupStore: env.BACKUPS });`);
+
+        expect(discover()).toStrictEqual(["0 3 * * *"]);
+    });
+
+    it("reads the keys of a crons handler map, in source order after backupCron", () => {
+        expect.assertions(1);
+
+        writeAt(
+            "src/server/index.ts",
+            `export default createWorker({
+                backupCron: "0 3 * * *",
+                crons: { "*/15 * * * *": async () => {}, "0 0 1 * *": rollUp },
+            });`,
+        );
+
+        expect(discover()).toStrictEqual(["0 3 * * *", "*/15 * * * *", "0 0 1 * *"]);
+    });
+
+    it("reads a backtick-quoted backupCron", () => {
+        expect.assertions(1);
+
+        // A `crons` KEY has no such form — a bare template literal is not a legal
+        // property name — so backticks only ever reach the `backupCron` read.
+        writeAt("src/index.ts", "export default createWorker({ backupCron: `0 3 * * *` });");
+
+        expect(discover()).toStrictEqual(["0 3 * * *"]);
+    });
+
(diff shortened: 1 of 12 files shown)


prisis:
Rebased onto current `alpha` (was 144 commits behind) and collapsed from 15 commits to 1.

Fourteen of the fifteen had already landed, several of them in a **later and better** form than this branch carried — so replaying them would have been a regression, not a rebase:

- `fix(scheduler): mint job ids the workflow engine accepts` — `alpha` now *refuses* a bad id with `400 INVALID_SCHEDULE_ID` rather than silently minting over it. Minting over a caller's id means two calls naming the same bad id run the job twice instead of the second getting a `409`. Strictly better, leading-dash fix included.
- `fix(config): keep cron triggers the app hand-wrote` and the four commits after it — these build the **JSONC comment marker** ownership record (`// lunora:crons [...]` above `triggers.crons`). `alpha` has the design that replaced it: the record lives in `package.json` under `lunora.crons`, with explicit degradation when the record is damaged. Reapplying the marker would have reverted that.
- `fix(scheduler): drop the originUrl nobody reads` — landed in #629.
- `fix(runtime): test the cron surface for emptiness, not presence` — already on `alpha`, tests and all. (Worth keeping visible: codegen emits `cronJobs: LUNORA_CRONS` unconditionally and that map is `{}` for a cron-free app, so the old presence check was true for every app and silently dropped the framework host's own `scheduled`.)
- `fix(scheduler): name the env var the DO really dispatches from` — `alpha`'s scheduler docs already say `env.LUNORA_ORIGIN_URL` is the dispatch target and that a caller-supplied one would be an SSRF vector.

What survives is the actual feature: **`feat(codegen): discover the crons configured on createWorker`** — the static `backupCron` / `crons` keys on a worker entry are discovered and added to the wrangler trigger set. One conflict on the way in, in `run-codegen.test.ts`, purely adjacent additions (`alpha`'s vector-introspector test and this branch's cron test land at the same spot); both kept.

Two things I checked rather than assumed:
- `packages/vite/__tests__/cron-sync.test.ts` looks missing on `alpha`, but #629 **deliberately deleted** it and moved that coverage into `packages/config/__tests__/reconcile-crons.test.ts` (+255 lines). Not restored.
- I thought the scheduler docs had gone stale showing `originUrl` on `createScheduler` examples. They have not — that was an artifact of my own half-applied cherry-pick. On a clean tree the only `originUrl` left is inside `httpDispatcher({ … })`, where it is still a real option.

Gates green: `build`, `test`, `lint:types`, `lint:eslint`, `lint:prettier`, `api:check`, `lint:package-json`, `check-generated-files`.

PR #639, opened 2026-09-06
fix(payment): key the mutating calls the providers let us key

prisis:
Follow-up to #633, which corrected `idempotency.ts`'s claim that a Polar refund was "THE ONE EXCEPTION" to every mutating provider call carrying a stable key. It is not, and that PR deliberately left the fixes here because whether a provider accepts a key is per-provider work.

## Stripe: two un-keyed calls, both fixed

`subscriptions.update` accepts a key as its third argument — one call site already passed one, two did not:

- **resume** now uses `resume_subscription:stripe:<subId>`, deliberately its own operation name. Reusing the cancel key on the same subscription would make a resume replay the cancel's cached response.
- **plan change** now uses `update_subscription:stripe:<subId>:<priceId>:<quantity>`. The target is in the key for the same reason `capturePayment` puts the amount in its own: one key across two parameter sets is a Stripe mismatch error, while an identical retry must still replay. This is the path that moves money immediately through proration.

`SubscriptionPatch.idempotencyKey` and a `resumeSubscription` options argument give callers the same override that cancel and capture already have — because a stable key has a real ceiling: re-issuing a target that was applied and then changed away from, inside Stripe's 24h window, replays instead of acting. Both are documented Stripe-only, and both types are `@experimental`, so no snapshot drift.

## Dodo: it cannot carry a key, and no header was guessed

The SDK builds an idempotency header only when `this.idempotencyHeader` is truthy. That field is `protected`, declared, read at `client.js:457-461` — and **assigned nowhere in the package**. `ClientOptions` exposes no such field either, so a subclass is the only route, and that still needs a header name Dodo honours. Its OpenAPI-derived resources declare idempotency **only as body fields on unrelated endpoints**; `CustomerCreateParams` and `SubscriptionChangePlanParams` carry none, and the API reference documents no request header.

So the key Lunora already passes on `customers.create` type-checks and is silently dropped. Proven against the real client with an injected fetch — 11 headers captured, none of them idempotency, so the assertion is not vacuous.

**No header was invented.** A wrong header name is worse than a documented gap, because it looks like protection. Instead there is a canary test: it drives the real SDK, asserts no idempotency header leaves, passes today by design, and **fails when a future release starts honouring the key** — pointing at the docblock that would then be stale.

## Creem, Polar, Autumn: checked, and also cannot

Creem's `upgrade` takes Speakeasy's `RequestOptions` (`timeoutMs`, `retries`, `retryCodes`, `serverURL`, `fetchOptions` — no idempotency field). Polar's `subscriptions.update` has the identical shape. `autumn-js` contains no idempotency surface at all.

So a prorated plan change is un-deduped on every provider except Stripe. `idempotency.ts` now names all three uncovered classes with the reason each cannot carry a key, and `docs/index.mdx` says "wherever the provider accepts one" rather than implying universal coverage.

## Verification

Four Stripe tests written and run against unfixed code first, each failing `expected undefined to be '…'` — no third argument reached the SDK. After: 28 pass. Package suite 256 pass.

All seven repo gates green, plus `check-generated-files.mjs`.

The over-refund and local-marker guards these keys complement are untouched — for Polar they remain the only protection, which is what `idempotency.ts` says.

## Summary by CodeRabbit

* **New Features**
  * Added optional idempotency key overrides when resuming or updating subscriptions.
  * Stripe subscription operations now use stable idempotency keys by default to help prevent duplicate actions during retries.
  * Documented that custom idempotency key overrides are honored only by Stripe.

* **Documentation**
  * Clarified idempotency coverage, replay behavior, and provider-specific retry handling.

### DIFF
--- a/packages/payment/__tests__/providers/dodopayments.test.ts
+++ b/packages/payment/__tests__/providers/dodopayments.test.ts
@@ -1,5 +1,6 @@
 import { createHmac } from "node:crypto";
 
+import DodoPayments from "dodopayments";
 import { describe, expect, it } from "vitest";
 
 import type { DodoPaymentsClientLike } from "../../src/providers/dodopayments";
@@ -240,6 +241,32 @@ describe("dodopayments adapter", () => {
 
         expect(typeof options?.idempotencyKey).toBe("string");
         expect(options?.idempotencyKey).not.toHaveLength(0);
+    });
+
+    it("pins that the SDK drops that key rather than sending it (the fake above cannot see this)", async () => {
+        expect.assertions(2);
+
+        let sent: Headers | undefined;
+        // The real client, not the structural fake: `buildHeaders` only emits an idempotency header
+        // when `this.idempotencyHeader` is truthy, and that field is declared `protected` and never
+        // assigned anywhere in the package — so the key we pass type-checks and goes nowhere. Flip
+        // this test to assert the header when a future SDK release starts setting it, and update the
+        // `@lunora/payment` idempotency docblock with it.
+        const client = new DodoPayments({
+            bearerToken: "test-key",
+            environment: "test_mode",
+            fetch: async (_input, init) => {
+                sent = new Headers(init?.headers);
+
+                return Response.json({ customer_id: "cus_1", email: "a@b.test" });
+            },
+            maxRetries: 0,
+        });
+
+        await client.customers.create({ email: "a@b.test", name: "user_1" }, { idempotencyKey: "customer:dodopayments:user_1" });
+
+        expect(sent).toBeDefined();
+        expect([...(sent as Headers).keys()].filter((name) => name.includes("idempotency"))).toStrictEqual([]);
     });
 
     it("ingests usage as a Dodo usage-event keyed on the customer id", async () => {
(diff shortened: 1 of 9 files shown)


prisis:
Rebased onto current `alpha` (was 112 commits behind).

One conflict, in the `idempotency` module docblock. `alpha`'s copy had been rewritten in the meantime and is the more detailed of the two, so I kept it and folded this branch's additions into it rather than the reverse. Two notes on the resolution:

- `alpha`'s docblock ended with *"**Two Stripe calls are un-keyed but trivially keyable** — `subscriptions.update` in `resumeSubscription` and in `updateSubscription` … Adding one is a follow-up, not a change to make blind."* That follow-up is this branch's first commit, so the paragraph was stale the moment it rebased. Replaced with the key-stability guidance it needs instead — that a key is stable for the logical *operation*, the resulting replay-vs-prorate toggle, and the explicit-override escape hatch.
- `alpha`'s per-provider breakdown of what is and is not keyable is strictly richer than this branch's, and already covers the Dodo inert-header finding, so that side was kept verbatim.

Gates green on the rebase: `build`, `test` (payment 269/269), `lint:types`, `lint:eslint`, `lint:prettier`, `api:check`, `lint:package-json`, `check-generated-files`.

PR #640, opened 2026-09-06
fix(dispatch): encode ctx.run args once, at the runner

prisis:
> **Stacked on #634** (`fix/r32-workpool-wire`) — based on that branch, not `alpha`. Merge #634 first.

`createDispatchRunner` posted `args` with no `encodeWire` against a shard that decodes unconditionally, so `ctx.run(fn, { n: 1n })` threw and a `Date` arrived as an ISO string.

## Six producers, not three

Every non-test caller of the runner:

| Site | Encoded before? |
|---|---|
| `scheduler/queue-workpool.ts` (`httpDispatcher`) | **yes** — #634 brackets it at `enqueue` |
| `queue/run-context.ts` — `ctx.run` in a push handler | no |
| `workflow/run-context.ts` — `ctx.run` in a workflow body | no |
| `agent/resolve-run.ts` | no |
| `agent/voice-do.ts` | no |
| `agent/create-agent-context.ts` | no |

Exactly one decode on the path: `shard-do.ts:5925` inside `runHandler`. The other `decodeWire` calls there are the WebSocket envelope routes and `decodeAdminArgs`.

## Why this encodes in the shared runner rather than per-package wrappers

The plan was five per-package `run` wrappers, to avoid double-encoding the one producer #634 already brackets. The runner turned out to be the better seam, for two reasons:

**The unsafe seam is structurally unreachable.** `handleSchedulerDispatch` takes the `create({ params })` branch only when the body carries a top-level string `workflow` field, and the runner's body literal is `{ args, functionPath, id, shardKey }` — there is no `workflow` key in it, ever. That is a property of the body shape, so it holds for all six producers at once instead of needing a per-package argument. (#634 established this for the queue path specifically; it generalises.)

**The failure modes are asymmetric.** With wrappers, a seventh producer that forgets one silently corrupts `Date`s — `decodeWire` is not idempotent, and a missed encode is indistinguishable from success until someone inspects a value. With one encode plus an `argsAlreadyEncoded` opt-out, the only way to get it wrong is to add a producer with its own serialising hop *and* forget to declare it — and the single existing such site is pinned by a test that fails eight ways if the flag is dropped.

## Verification

Regression tests in `dispatch`, `queue`, `workflow` and `agent`, each written and run against unfixed code first — `TypeError: Do not know how to serialize a BigInt` at `create-dispatch-runner.ts:306`, and `expected '2026-06-01T12:00:00.000Z' to be an instance of Date`. Each asserts the decoded **type**, not merely the absence of a throw, plus a plain-JSON identity case.

The no-double-encode property is pinned on #634's own suite: the POSTed `args` must be byte-identical to what `enqueue` put on the queue. Proven to fire by deleting the opt-out — 8 failures, including `expected [ '$lunora.wire$', 'bigint', '7' ] to be 7n`.

All repo gates green including `check-generated-files.mjs`, `no-nul-bytes`, and `test:coverage` for the five touched packages. Verified the fix reaches the bundle rather than only the source.

`packages/queue/tsconfig.json` drops `outDir`/`rootDir` — the documented convention for importing `shared/*`, mirroring `@lunora/workflow`, which carries the same comment. Six lint findings in my own files were fixed in source, none suppressed.

## Found while doing this, deliberately not fixed

**`ctx.run`'s return value has the mirror-image gap.** The shard encodes its result (`shard-do.ts:6047`) and wraps it as `{ result }`; the runner ends with a bare `return JSON.parse(text)` — no unwrap, no decode. So `await ctx.run(fn)` resolves the *envelope*, wire-encoded: a mutation returning a `bigint` hands the caller `{ result: ["$lunora.wire$","bigint","…"] }`, and `agent-loop.ts` casts that straight to `AgentMessageRow[]`.

There is no end-to-end test of a `ctx.run` return value anywhere, which is why it has stayed invisible. Fixing it changes `ctx.run`'s public return shape, so it belongs in its own change rather than riding along here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Dispatch arguments now reliably preserve supported values such as large integers, dates, and binary data when delivered to handlers.
  - Missing arguments are normalized to an empty object, improving consistency for dispatched jobs.
  - Arguments that are already prepared for delivery are no longer encoded twice, preventing potential processing errors.

- **Documentation**
  - Updated dispatch and scheduler documentation to clarify argument forwarding and supported delivery behavior.

### DIFF
--- a/packages/agent/__tests__/resolve-run-args-wire.test.ts
+++ b/packages/agent/__tests__/resolve-run-args-wire.test.ts
@@ -1,0 +1,69 @@
+/**
+ * The agent loop's dispatcher is `@lunora/dispatch`'s runner (built here by
+ * `resolveAgentRun`, and identically by `createAgentContext` and the voice DO's
+ * `resolveRun`), POSTing to `/_lunora/scheduler/dispatch`. The shard's dispatch
+ * loop decodes the body's `args` exactly once
+ * (`decodeWire(payload.args ?? {})`), so this end has to encode: without it a
+ * `bigint` throws in `JSON.stringify` before the request leaves, and a `Date`
+ * reaches the handler as an ISO string.
+ */
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { decodeWire } from "../../../shared/wire-codec";
+import resolveAgentRun from "../src/resolve-run";
+import type { AgentFunctionReference, AgentRunFunction } from "../src/types";
+
+const ENV = { LUNORA_ADMIN_TOKEN: "admin-token", LUNORA_ORIGIN_URL: "https://app.example" };
+const REF: AgentFunctionReference = { __lunoraRef: "agents:agentAppendMessage" };
+
+/** Never reached — the owner branch always builds a fresh dispatcher. */
+const contextRun: AgentRunFunction = async () => undefined;
+
+/** Dispatch `args` through the owner-scoped agent runner and hand back what the shard's single decode gives the handler. */
+const handlerArgs = async (args: Record<string, unknown>): Promise<Record<string, unknown>> => {
+    let body = "";
+
+    vi.stubGlobal("fetch", async (_url: string, init: RequestInit) => {
+        body = init.body as string;
+
+        return new Response(null, { status: 200 });
+    });
+
+    await resolveAgentRun(contextRun, "user-a", ENV)(REF, args);
+
+    return decodeWire((JSON.parse(body) as { args?: unknown }).args ?? {}) as Record<string, unknown>;
+};
+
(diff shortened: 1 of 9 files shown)


prisis:
Superseded. The source fix is already on `alpha`: `create-dispatch-runner.ts` encodes args at the runner, `queue-workpool.ts` encodes at the producer, and the `argsAlreadyEncoded` opt-out this branch introduced landed with #643 (which repaired `alpha` after those two halves merged separately and double-encoded).

What had **not** landed is the four args-wire test files. They are ported in #645, with the dispatch-local `argsAlreadyEncoded` pin kept — verified to fail if the flag honouring is removed from the runner.

PR #641, opened 2026-09-06
fix(dispatch): unwrap and decode what ctx.run returns

prisis:
> **Stacked on #640** (`fix/r32-ctx-run-wire`), which fixes the args direction. Merge #640 first.

The mirror image of #640, and the more severe half: **every durable agent turn threw at its first history read.**

## The defect

The shard encodes its result (`shard-do.ts:6047`, `buildDispatchResponse(mutatorClass, encodeWire(result))`) and wraps it as `{ result }` or `{ lastMutationId, result }`. `dispatchToShard` and `handleSchedulerDispatch` forward that verbatim, and `createDispatchRunner` ended with a bare `return JSON.parse(text)` — no unwrap, no decode. So `await ctx.run(fn)` resolved the **transport envelope, wire-encoded**.

The sibling dispatcher behind an `httpAction`'s `ctx.runQuery`/`ctx.runMutation` has always done this correctly — `create-worker.ts:3721` is `return decodeWire(payload.result) as R`. The dispatch runner was the one seam that disagreed, which is the same shape as every other codec gap found in this campaign.

## It was live, not latent

`agent-loop.ts:697` is `const rawHistory = (await run(listMessages, { key: threadKey })) as AgentMessageRow[]`, and `workflow.ts:113` wires `run` to `resolveAgentRun(context.run, …)` — both branches of which are `createDispatchRunner`. Run against the real built exports with the pre-fix envelope:

```
envelope (pre-fix) / buildModelMessages -> THROWS TypeError: options.history is not iterable
envelope (pre-fix) / splitForCompaction -> THROWS TypeError: history.slice is not a function
rows     (post-fix) / buildModelMessages -> OK  (2 model messages)
```

Every durable agent turn and every voice turn (`voice-turn.ts:338`) died there. Same class, silent rather than throwing: `bootstrap?.outcome`, `outcome?.dequeued` (so a queued run was never woken), `getState`, and `readRetrievedContext` (so every memory source returned no context).

## The change

`return decodeWire((body as { result?: unknown } | null)?.result)` — the same chokepoint #640 uses for the outbound direction, so the runner owns both halves of the wire, matching `mail/inbound/shard.ts` and `scheduler/do-client.ts`.

**No caller compensated**, so nothing had to be un-compensated and there is no double-unwrap risk. All six producers are unchanged; `voice-do.ts:529` already read `ensured?.outcome` — the unwrapped shape — and now actually gets it. Only three test fixtures asserting the old bare shape were updated.

`lastMutationId` and `commitCursor` are dropped deliberately: both are client-session bookkeeping feeding `onMutationAck`/`onCommitCursor`, and the runner has no session and no subscriptions. Grepping both keys across the five consuming packages found no reader.

**BREAKING CHANGE**, noted in the commit body — though `api:check` shows **no drift**, because `DispatchRunFunction` is `Promise<unknown>` either way. The behaviour changed without the type surface moving, which is precisely why nothing caught it.

## Verification

The finding rests on there being **no end-to-end test of a `ctx.run` return value anywhere**, so that test came first: 9 cases in `dispatch-result-wire.test.ts`, **8 failing** against unfixed code — `expected 'undefined' to be 'bigint'`, `expected undefined to be an instance of Date`, `expected { result: {…} } to strictly equal { count: 3 }`, and both envelope variants. After: 9/9.

`shard-do.dispatch-result-wire.test.ts` pins the shard half against a real `ShardDO` over `node:sqlite`, including a dedup replay returning the identical encoded envelope. `shard-do.client-watermark.test.ts` is untouched and still passes.

All repo gates green including `check-generated-files.mjs` and `no-nul-bytes`.

Three assumptions were corrected by running rather than reading: `encodeWire(undefined)` tags rather than drops the key; a custom-mutator replay hits the watermark ack, not the idempotency cache, and is unreachable from `ctx.run` anyway; and a re-fired `ctx.run` gets `{ commitCursor, result }` because `dedupId` puts it on the mutation path of a CDC-enabled shard.

### DIFF
--- a/packages/agent/__tests__/resolve-run.test.ts
+++ b/packages/agent/__tests__/resolve-run.test.ts
@@ -31,7 +31,9 @@ describe(resolveAgentRun, () => {
             capturedHeaders = init.headers as Record<string, string>;
             capturedBody = init.body as string;
 
-            return Response.json([{ content: "hi" }]);
+            // The shard's real transport envelope — the loop's `listMessages`
+            // read gets the ARRAY back only because the runner unwraps it.
+            return Response.json({ result: [{ content: "hi" }] });
         });
 
         vi.stubGlobal("fetch", fetchSpy);
@@ -52,7 +54,8 @@ describe(resolveAgentRun, () => {
         expect(capturedHeaders["x-lunora-userid"]).toBe("user-a");
         expect(JSON.parse(capturedBody)).toStrictEqual({ args: { key: "thread-1" }, functionPath: "agents:agentMessages" });
 
-        // The dispatcher resolves the function's JSON return value.
+        // The dispatcher resolves the function's own return value, unwrapped out
+        // of the `{ result }` envelope — not the envelope.
         expect(result).toStrictEqual([{ content: "hi" }]);
     });
 });
--- a/packages/dispatch/__tests__/create-dispatch-runner.test.ts
+++ b/packages/dispatch/__tests__/create-dispatch-runner.test.ts
@@ -11,7 +11,9 @@ describe("createDispatchRunner", () => {
     it("pOSTs to /_lunora/scheduler/dispatch with the bearer + envelope and resolves the JSON body", async () => {
         expect.assertions(5);
 
-        const fetchImpl = vi.fn<typeof fetch>(async () => Response.json({ ok: 1 }, { status: 200 }));
+        // The stub answers the shard's real transport envelope; the runner
+        // unwraps `result` and decodes it (see `dispatch-result-wire.test.ts`).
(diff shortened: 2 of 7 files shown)


prisis:
Superseded. The source fix is already on `alpha` — `create-dispatch-runner.ts:412` is `return decodeWire((parsed as { result?: unknown }).result);`, with the envelope shape checked just above it.

The two result-wire test files are ported in #645, with one assertion changed. This branch asserted that a 200 of `{"commitCursor":7}` should resolve to `undefined`. It should not: all three arms of `buildDispatchResponse` spell `result` literally, so a 200 without the key is a body the shard cannot emit — something else answered it. Reading it as `undefined` would hand a handler a silent wrong answer, so #645 pins the refusal instead.

PR #648, opened 2026-09-07
security(server): govern .global() writes and scope shape read policies

prisis:
First of three. **Base `alpha`.** The chain: this → #PR2 → #PR3.

Closes six authorization holes found by a core-package hunt, plus the three
follow-ups a review of this branch turned up — including a fail-open this
branch's own shape-scoping change introduces, which is why they land together
rather than in a later PR.

## `.global()` rows addressed by id were ungoverned by RLS write policies

`lookupById` is a shard-local seam and returns `null` for a global row. Its
docblock promised *"the caller keeps its probe fallback"* — `locateRow` had no
fallback, because the probe path ran only when `lookupById` was **absent**, and
the real shard writer always has it. `gateById` then hit
`return perform(base)` under the comment *"the id is in no policy-gated table"*,
which for a global row is false: it **is** gated.

With `update: () => false` and `delete: () => false` on a global table:

```
before   patch/delete/deleteAll → RESOLVED
         global writer saw 4 ops · rows left: 0   (destroyed)
after    all four → THREW FORBIDDEN
         global writer saw []   · rows left: 2   (intact)
```

Affects `patch`, `replace`, `delete`, `deleteMany`, `restore`, `deleteAll`,
`hardDelete` and the facade `upsert`'s patch step. `insert` was never affected —
it is table-addressed. `ctx.db.get()` returned `null` for the same rows under
both `rls()` and `mask()`; it now returns the row, subject to policy.

The existing suite passed because `rls.test.ts:1232` uses a fake writer with no
`lookupById` — it exercises the probe path production never takes.

## Live subscription shapes took the OR of every procedure's read policies

The registry was built from `Object.values(LUNORA_FUNCTIONS)` — every function
naming the table, internal ones included — while request-time scope is
per-procedure. One group granting unconditionally unrestricted the shape:

```
tenant procedure alone      → {"ownerId":"u1"}
+ one admin-list procedure  → {}      every row to every socket
```

Now scoped per shape via `defineShape({ use: [guard] })`.

**Breaking:** a shape no longer inherits project-wide policies.

## …and that scoping introduced a fail-open, fixed here

Under `.rls("required")` a shape with no `use` denies — fail-closed. Under the
**opt-in default** it replicated *unfiltered*: an app whose table is tenant-scoped
by `rls()` on its queries plus a `defineShape` with no `use` went from
`{AND: [{channelId}, {tenantId}]}` to `{channelId}` — silently, no error.

The generated DO now refuses to boot, naming every offending shape and its
table, when the project's own read policies govern a table a shape replicates
without declaring `use`. `use: []` is the explicit *"this shape is deliberately
ungoverned"* opt-out. Skipped under `required`, where omission already denies.

Emitted only when a project has both shapes and read policies, so a shape-free
or policy-free project's `shard.ts` is byte-identical.

## Also

- **A second `rls()` step dropped the first's filter** under `.rls("required")`
  (`["mine"]` → `["mine","theirs"]`). `RLS_UNWRAP_SYMBOL` is now non-enumerable
  so it cannot ride a spread, and the wrapper carries its accumulated chain;
  reads AND-compose across steps, and a write must be allowed by every step
  gating its table.
- **A middleware that never calls `next()`** skipped every later `.use()` —
  `rls`, `mask`, `storageRules` — while the handler still ran on the unwrapped
  `ctx.db`, and `fn.rls` was still hoisted so studio believed it guarded. Now a
  clear error.
- **`count`/`rank` failed closed under an allow-all policy.** The test is now
  whether the computed base *narrows*, matching `mergeBaseWhere`.
- **The A1 probe fallback** cost `2 + N` subrequests per row on a chunked
  `deleteAll` over global tables — ~960 for 80 rows, past workerd's 1000
  ceiling mid-erase. Both callers know the table, so the probe scope is now
  separate from `expectedTable` (which must stay unpinned for the global
  fallback): `2 + N` → `3`.
- `requireUnrestrictedReadBase` could only ever return `undefined` while three
  call sites still threaded it into `mergeBaseWhere`. Now an assertion; −15 lines.
- Docblocks that asserted the opposite of the code, rewritten. `use` is
  documented on the docs site for the first time, and the RLS caution — which
  said only *roles* fail to reach shapes — corrected.

## Not changed, deliberately

`deleteMany`'s `deleted` count. It looked like an over-report, but
`packages/server/src/types.ts:1170` documents it as the number of ids
**requested**, and the unwrapped writer matches. Changing only the RLS path
would make the same call report different numbers with and without a policy.

## Verification

Every regression test was run against the unfixed code first and observed to
fail. Gates green with the cache disabled: `vis run test --no-cache` (175),
`lint:types`, `lint:eslint`, `lint:prettier`, `api:check`, `lint:package-json`,
`check-generated-files`.

## Summary by CodeRabbit

* **New Features**
  * Shapes can opt into specific row-level security policies for replication and reads.
  * Multiple row-level security middleware steps can be combined, with all applicable policies enforced.
  * Added validation to prevent startup when governed shapes omit required read-policy declarations.

* **Bug Fixes**
  * Fixed global records incorrectly appearing unavailable inside masked procedures.
  * Improved count and ranking operations when policies do not restrict returned rows.
  * Middleware now reports an error when a step does not call `next()`.

* **Documentation**
  * Clarified middleware usage, policy behavior, and database lookup semantics.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -23,6 +23,10 @@
             # Audit-round fix PRs are stacked on this branch so each group is
             # reviewable on its own; auto review skips any base not listed here.
             - "fix/audit-round-5"
+            # Round-33 core-defect chain: the authorization group lands first,
+            # carrying the fix for the fail-open its own shape-scoping change
+            # introduces, and the rest is based on it.
+            - "fix/r33-server-authz-complete"
             # The audit chain: each group PR is based on the previous one.
             - "fix/r7-auth-agent-mail"
             - "fix/r7-payment"
--- a/api-snapshots/lunora.api.md
+++ b/api-snapshots/lunora.api.md
@@ -741,6 +741,10 @@ Re-exported from `@lunora/server` — signature tracked at its source.
 
 Re-exported from `@lunora/server` — signature tracked at its source.
 
+### `ShapeGuardDeclaration` (interface)
+
+Re-exported from `@lunora/server` — signature tracked at its source.
+
 ### `ShapeReadWhereRequest` (interface)
 
 Re-exported from `@lunora/server` — signature tracked at its source.
@@ -1070,6 +1074,10 @@ Re-exported from `@lunora/server` — signature tracked at its source.
 Re-exported from `@lunora/server` — signature tracked at its source.
 
 ### `asBucketStorage` (const)
+
+Re-exported from `@lunora/server` — signature tracked at its source.
+
+### `assertShapesDeclareReadPolicies` (const)
 
 Re-exported from `@lunora/server` — signature tracked at its source.
 
@@ -5471,6 +5479,10 @@ Re-exported from `@lunora/server` — signature tracked at its source.
 
 Re-exported from `@lunora/server` — signature tracked at its source.
 
+### `ShapeGuardDeclaration` (interface)
+
+Re-exported from `@lunora/server` — signature tracked at its source.
+
 ### `ShapeReadWhereRequest` (interface)
 
 Re-exported from `@lunora/server` — signature tracked at its source.
@@ -5800,6 +5812,10 @@ Re-exported from `@lunora/server` — signature tracked at its source.
 Re-exported from `@lunora/server` — signature tracked at its source.
 
 ### `asBucketStorage` (const)
+
+Re-exported from `@lunora/server` — signature tracked at its source.
+
+### `assertShapesDeclareReadPolicies` (const)
 
 Re-exported from `@lunora/server` — signature tracked at its source.
 
--- a/api-snapshots/server.api.md
+++ b/api-snapshots/server.api.md
@@ -1976,6 +1976,15 @@ interface ShapeDefinition<Args extends ValidatorMap = ValidatorMap, Context = Qu
 }
 ```
 
+### `ShapeGuardDeclaration` (interface)
+
+```ts
+interface ShapeGuardDeclaration {
+    readonly table: string;
+    readonly use?: unknown;
+}
+```
+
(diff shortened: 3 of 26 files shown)


prisis:
CodSpeed's report on this branch overstates the regression by ~3×, and two of its five entries are in packages this PR does not touch. Measured locally instead, same machine, on the repo's own `rls-overhead` bench:

| | `rls(true)` wrapper install |
|---|---|
| `alpha` | 902,611 hz |
| this branch, first push | 747,651 hz — **−17%** |
| this branch, now | 794,031 hz — **−10%** |

CodSpeed reported −52.95% for the same benchmark, alongside its own "different runtime environments detected" warning — the signature of head and base being measured under different conditions rather than a 2× slowdown.

The real cost was `wrapDatabase` building the gated-table set by flattening every step's `perTable` keys through an intermediate array into a fresh Set, on **every wrapper install** — i.e. every guarded query. Chaining made it take the whole step list, but one step is the overwhelming majority of chains and its `perTable` is already a keyed Map, so that case now reuses it (`b9cf565`).

The residual ~10% is the chain-aware read path, the `narrowsReads` check and the probe scoping. That is spread across the wrapper's methods rather than sitting in one hotspot, and it is the price of the authorization fixes this PR exists for — a bypass that let `patch`/`delete`/`deleteAll` run against `.global()` rows with an explicitly denying policy. I'd take 10% on the guarded-query path for that; flagging it explicitly rather than letting the CodSpeed number stand as the record.

The other two regressed benchmarks (`packages/do`'s `normalizeLogFields`, `packages/values`' `parse-hotpath`) are in code this PR does not modify.

Issue #651, opened 2026-09-07
codegen silently drops a spread in .input() and a factory-assigned export — exit 0, no diagnostic


prisis:
## Versions

| package | version |
| --- | --- |
| `@lunora/cli` | `1.0.0-alpha.233` |
| `lunorash` | `1.0.0-alpha.236` |
| `@lunora/server` | `1.0.0-alpha.104` |

## Summary

`lunora codegen` registers an export only when the export's initializer is *literally* a builder chain, and it reads the `.input()` object literal *syntactically*. Two ordinary constructs are therefore dropped in silence — codegen exits 0 and prints nothing:

1. a spread inside `.input()` loses the spread fields
2. an export assigned from a factory call is not registered at all

## Reproduction

`lunora/probe.ts`:

```ts
import { internalQuery } from "./_generated/server";
import { v } from "lunorash/server";

// (1) spread in .input()
const sharedArgs = { alpha: v.string(), beta: v.optional(v.number()) };
export const probeSpread = internalQuery
    .input({ ...sharedArgs, gamma: v.boolean() })
    .query(async ({ args }) => `${args.alpha}${args.gamma}`);

// (2) export assigned from a factory call
const makeProc = () => internalQuery.input({ delta: v.string() }).query(async ({ args }) => args.delta);
export const probeFactory = makeProc();
```

```console
$ lunora codegen
$ echo $?
0
$ grep probe lunora/_generated/api.ts
        probeSpread: FunctionReference<"query", { gamma: boolean }, string>;
```

## Expected vs actual

|  | expected | actual |
| --- | --- | --- |
| `probeSpread` args | `{ alpha: string; beta?: number; gamma: boolean }` | `{ gamma: boolean }` |
| `probeFactory` | present in `api.ts` | **absent entirely** |
| exit code / diagnostic | non-zero, or a warning naming the export | `0`, silence |

The runtime validator still enforces the real `.input()`, so the generated reference and the deployed function disagree in both directions: a caller passing `alpha` is told the property does not exist, and a caller that satisfies the generated type fails validation at runtime.

## Why this is expensive

Every instance surfaces as a *name* error somewhere else entirely, so it reads as "you spelled something wrong" rather than "your function was dropped". Porting a 92-table Convex backend we hit five variants of the same root cause:

| form | how it surfaced |
| --- | --- |
| `export default someProcedure` | `Property 'triggers_execute' does not exist` at the caller |
[...]

PR #676, opened 2026-09-08
fix(codegen): report when the procedure type check cannot run

prisis:
Closes the factory half of #651 — by fixing the reason the existing check was silent, not
the drop itself.

## What was actually wrong

`procedure_not_registered` already existed (codegen@1.0.0-alpha.137) and already fired on
every variant of `export const x = factory()` I could reproduce in `examples/todo-app`:
cross-file factories, factories with parameters, `.output(v.number())`, `.use()`
middleware, nested directories. So the drop was reported. What wasn't reported was the
check failing to run: it is a **type**-level check, and where types don't resolve for
`lunora/`, every binding is `any`, nothing matches, and codegen prints exactly the silence
it prints for a clean project. That is the indistinguishability the reporter identified as
the real cost — "a broken probe and a working repro are indistinguishable from the
outside."

## Changes

**`procedure_type_check_unavailable` (WARN).** Probes the checker against procedures
codegen *did* register and reports when not one of them resolves. It names the witness it
could not read, so the claim is one `tsc` invocation from being reproduced:

```
[WARN] procedure_type_check_unavailable — Type resolution for `lunora/` is unavailable —
the registered procedure `list` in `todos` types as `any`, as does every other one — so
codegen cannot tell you when an export is dropped from `_generated/api.ts`.
      ↳ Check what codegen sees: `list` should have a procedure type, not `any`. …
```

Fires only when **every** witness fails, so the verdict doesn't depend on file order in a
partially-resolving project. Witnesses are restricted to builder terminals — the bare
factory form `query({ args, handler })` types as `any` even in a healthy install, so
taking it as a witness blamed the toolchain for a call that is simply broken. Filtered out
of the emitted `_generated/shard.ts`, which is committed: a fact about the build machine
must not land in a tracked file.

**The type table was not exhaustive.** `RegisteredStream` was missing (a dropped
`.stream()` procedure was never reported, and a stream-only app got told to fix a
perfectly good `tsconfig.json`), and so were `RegisteredLifecycleHook` and
`RegisteredReactor` — a factory-assigned `onConnect` / `onDisconnect` / `onShardInit` /
`onQueryChange` was dropped in total silence. That is worse than the reported case: a
dropped query eventually surfaces as `Property 'x' does not exist` at a call site, while a
dropped hook has no caller at all, so it simply never fires.

**And it was never only procedures.** Seven sibling discoverers share the same shape: each
walks exported declarations and `continue`s past anything whose initializer is not literally
its own `define*` call — `mutators.ts:153`, `shapes.ts:146`, `workflows.ts:208`,
`queues.ts:131`, `agents.ts:144`, `containers.ts:281`, `migrations.ts:132`. All seven now
report, each with its own registering call in the remediation.

Two things the type table alone cannot do, both learned by running the check against a real
install rather than a hand-built fixture:

- A row without the kind's identity in `Registrations` reports every **healthy** export of
  that kind as dropped. The IRs disagree about identity — procedures, mutators, shapes and
  migrations carry a `filePath`; workflows, queues, agents and containers record only an
  `exportName` — so the diff takes both keyings. Crons stay out: `CronJobIR` records no
  exporting binding to match at all.
- Six kinds are read from **one module each** (`lunora/shapes.ts`, `lunora/mutators.ts`, …),
  so the same call anywhere else is skipped however directly it is assigned. Reporting the
  indirection cause there tells the reader to inline a factory that is not the problem — so
  a misplaced registration now says to move it, and says the declaration itself is fine.

`registration-types-parity.test.ts` locks the table: it scrapes every `Registered*` type in
the workspace, pins the four `*Definition` types the `define*` APIs return, and lists the
non-registrations (codegen's own emit metadata, `@lunora/mcp`'s internal tool bookkeeping)
with the reason each is excluded.

The hook remediation says to substitute the factory the author called rather than naming
one: `lifecycle` is typed as the whole `LifecycleEventKind` union, so the three hook
factories are one indistinguishable type. Only a reactor's is a literal, so `onQueryChange`
is named exactly.

**Unresolved types are rejected before the type name is read.** TypeScript keeps the alias
symbol of an unresolved annotation, so `const x: RegisteredQuery<…>` still answers
`"RegisteredQuery"` off an `any` — the check was string-matching the text the author typed,
which a factory fools exactly as easily as the syntactic scan does. Two pre-existing tests
were passing for that reason and now use fixtures that genuinely resolve.

**The remediation is stored per type rather than assembled from a kind.** `.stream()` is a
terminal on the QUERY builder, so a dropped stream needs `query.….stream(handler)`; every
kind used to be handed `query.….query(handler)`, telling an action author to write a query.

## Behaviour changes worth reviewing

- `procedure_not_registered` **stops** firing in projects where types don't resolve. Those
  projects now get the one honest finding instead of a coincidence-based hit.
- Dropped streams, lifecycle hooks, reactors, mutators, shapes, migrations, workflows,
  queues, agents and containers are all **newly** reported, as is a single-module kind
  registered from the wrong file.

## Verification

- `pnpm --filter "@lunora/codegen" run test` — 139 files, 1769 tests.
- `tsc --noEmit`, eslint, prettier clean. `pnpm run api:check` — 54/54 snapshots, fresh build.
- New `__tests__/discover/unregistered-procedures.test.ts` (5 tests), including the guard
  that the advisory stays quiet on a resolving project — the assertion that would have
  caught the `RegisteredStream` false positive.
- End-to-end with a rebuilt CLI against the real installed declarations of all ten kinds,
  in a harness app that actually declares `@lunora/workflow`, `@lunora/queue`,
  `@lunora/agent` and `@lunora/container`: every factory-produced registration reported
  with the right call, **zero** of the directly-assigned ones reported, a misplaced shape
  told to move, nothing in `shard.ts`; bare-factory form on a healthy install → no finding;
  tsconfig-less copy → the type-check diagnostic naming its witness.
- That harness is what caught the mistake worth naming: the first version of this change
  flagged six healthy registrations, because the unit fixture piled every kind into one
  module while discovery reads a specific module per kind. A hand-built fixture agrees with
  whatever the author assumed.


prisis:
Both CodeRabbit findings addressed in bc9efcdb2.

**The reactor remediation was invalid TypeScript — good catch, and fixed.** `onQueryChange`
takes two positional arguments (`packages/server/src/reactors.ts:150`), so the snippet the
finding told people to paste failed to compile. Verified against the real declarations
rather than taken on faith:

```
onQueryChange(async () => 1, async () => {})                      → tsc clean
onQueryChange({ select: async () => 1, handler: async () => {} }) → TS2554: Expected 2 arguments, but got 1
```

Worth recording why this got through: the unit-test stub declared `onQueryChange` as taking
a config object, so the fixture encoded the same wrong assumption as the code and the test
agreed with the bug. The stub now mirrors the real positional signature. A remediation
string is the one output whose whole job is to be pasted, so it is now checked by compiling
it, not by asserting it against a stub I wrote myself.

**On sharing the finding name (nitpick): taken, but not as an export.** This repo forbids
mixing a default export with named exports in one file (AGENTS.md), and
`unregistered-procedures.ts` is default-only like every sibling discoverer — so
`export const PROCEDURE_TYPE_CHECK_UNAVAILABLE` would trade one drift risk for a convention
break.

The underlying risk is real though: the filter silently ceasing to match, letting a
machine-specific finding back into the committed `_generated/shard.ts`. That is now locked
on the output instead, which catches strictly more than a shared constant would — a rename,
a filter deleted, or emit starting to embed advisories by another path:

```ts
expect(result.advisories.map((entry) => entry.name)).toContain("procedure_type_check_unavailable");
expect(result.generated.shard).not.toContain("procedure_type_check_unavailable");
```

139 files / 1765 tests, `tsc --noEmit`, eslint and `api:check` (54/54) all clean.

PR #677, opened 2026-09-09
fix(config,cli,vite): close the fail-open holes in the entry export check

### LINKED ISSUE #652: lunora verify reports "project is valid" for a project wrangler refuses to bundle (unexported Durable Object class)
## Versions

| package | version |
| --- | --- |
| `@lunora/cli` | `1.0.0-alpha.233` |
| `lunorash` | `1.0.0-alpha.236` |
| `@lunora/scheduler` | `1.0.0-alpha.57` |

## Summary

`lunora verify` is the cheap CI gate — *"Validate wrangler.jsonc + codegen dry-run + tsc --noEmit (no files written)"*. It reports **"project is valid"** and exits 0 for a project `wrangler` refuses to bundle, so a PR check goes green and the deploy fails.

## Reproduction

Comment out a Durable Object class re-export in the worker entrypoint, leaving the matching `durable_objects.bindings` entry in `wrangler.jsonc` in place:

```diff
-export { SchedulerDO } from "@lunora/scheduler";
+// export { SchedulerDO } from "@lunora/scheduler";
```

```console
$ lunora verify
  warn  verify: warnings:
  warn    - durable_objects.bindings declares the SchedulerDO but vars.LUNORA_ORIGIN_URL is unset — …
    ok  verify: project is valid (with warnings)
$ echo $?
0

$ lunora build
✘ [ERROR] Your Worker depends on the following Durable Objects, which are not exported in your entrypoint file: SchedulerDO.
$ echo $?
1
```

Note that `verify` *does* read that same binding — it warns about `LUNORA_ORIGIN_URL` on it (the check added for #625). So `durable_objects.bindings` is already parsed; it is just never cross-checked against the entrypoint's exported names.

## Impact

This is precisely the class of failure `verify` exists to catch, and the gap is invisible until the expensive command runs.

In our case the missing export would have taken down `ctx.scheduler.runAfter` / `runAt` wholesale — cron continuations, background work scheduled after each AI response, and every other `scheduleAfter` in the tree — with `verify` green throughout. Teams that run `verify` in PR checks and `build` only on the deploy job get the failure at the worst moment.

## Suggested fix

During `verify`, cross-check `durable_objects.bindings[].class_name` (and `workflows[].class_name`) against the entrypoint's exported names. That is a static check over files `verify` already parses — no bundle required — and it can name the missing export directly:

```
verify blocked: wrangler.jsonc declares durable_objects binding `SCHEDULER`
(class_name: SchedulerDO), but the entrypoint src/server.ts does not export
`SchedulerDO`. Add `export { SchedulerDO } from "@lunora/scheduler";`.
```

The alternative — having `verify` run `wrangler deploy --dry-run` the way `build` does — would also close it, but costs the thing that makes `verify` worth running separately.

prisis:
Follow-up to #653, which added this cross-check but left it switched off for most layouts. #652 has
stayed open across four re-tests because of that.

`lunora verify` reported **"project is valid"** and exited 0 for a project `wrangler` refuses to
bundle:

```console
$ lunora verify
    ok  verify: project is valid (with warnings)     exit=0
$ lunora build
✘ [ERROR] Your Worker depends on the following Durable Objects, which are not exported in your
          entrypoint file: SchedulerDO.                exit=1
```

The check (`collectUnexportedClassErrors`) was already there. It reported nothing whenever the
worker entry could not be resolved or read — and that was most of the time.

## The four bail-outs

| `main` | before | after |
| --- | --- | --- |
| adapter build output (`dist/_worker.js`, `.svelte-kit/…/_worker.js`) | no check | falls through to the authored-entry probe |
| absent (`@cloudflare/vite-plugin` supplies it) | no check | probes, and `src/server.ts` is now in the fallback list |
| any entry with a bare `export * from "./barrel"` | **whole check off** | relative stars followed, bounded + cycle-safe |
| `virtual:lunora/worker` (class-A) | no check | exports decided from the generator |

`src/server.ts` was missing from `WORKER_ENTRY_FALLBACKS` while already leading `lunora registry`'s
reconcile probe — the two lists had drifted. That is where the astro / solid-v2 / standalone
templates and `@lunora/astro`'s `serverEntry` compose.

The star case matters most: `@lunora/codegen` *itself* tells the entry to write
`export * from "./lunora/_generated/workflows.js"`, so a project that followed the generated
instruction had no cross-check at all.

Class-A gets its own remedy, because `@lunora/vite` generates that entry and there is no file to add
a re-export to. `SchedulerDO` / `SessionDO` there have no route at all — they are not `defineAgent` /
`defineContainer` / `defineWorkflow` declarations, so the first draft's "declare it in one of those"
was hours of dead end.

## Two more commands were green on the same trees

- **`lunora doctor`** read the *pure* validator, which cannot see the entry, so the cross-check never
  ran there either (the issue body says it did not). Now reports `wrangler-class-unexported` — its
  own code, not `declared-export-missing`: that one fires for the same container off inference, and
  `DOCTOR_CODES` is documented as the `--format json` contract a CI job branches on.
- **`lunora verify` had no `--env`.** `durable_objects` is not inheritable, so a binding declared only
  under `env.production` was never cross-checked, while `deploy --env production` would catch it. The
  cheap PR gate was validating a surface the deploy does not use.

## Fail-open, deliberately

This validator blocks deploys and `@lunora/vite` throws on its errors, so every undecidable case
reports nothing. A first draft of this change got that wrong in one place — probing the fallbacks for
*any* unreadable `main` resolved a hand-written `src/worker.js` (a supported entry) to an unrelated
`src/index.ts` and called its correctly exported `ShardDO` missing. Two guards fix it:

- the JS `main` gate is by **location**, not extension — a bundled `_worker.js` exports only the SSR
  handler, but `src/worker.js` is an ordinary entry, and both are `.js`;
- a **probed** entry is only trusted when it exports `default`. A module worker must; a client entry
  or barrel usually does not. (Checked: all 13 templates and 12 examples do.)

A `main` naming an absent source file reads nothing rather than diverting, and draws a warning of its
own — never an error, since a tree passes through that state mid-rename and `lunora dev` would stop.

## Also in here

`@lunora/config` now owns `GENERATED_CLASS_MODULES` / `GENERATED_DIRECTORY`, which existed in **four**
places (`@lunora/vite`, this validator, `reconcile-bindings`' type union, and a parity test that could
not catch drift in the direction that mattered — a stale copy fails *closed*). `@lunora/vite` depends
on `@lunora/config`, so config owning it is the direction the graph allows; vite re-exports both
names, so its public surface is unchanged. `UNEXPORTED_CLASS_MARKER` is exported because doctor keys
off it across packages and a copy-edit would otherwise disable that check silently.

## Verification

- 732 config / 1520 cli / 224 vite tests — **46 new**, one runnable case per hole and per fail-open
  guard, plus a fixture pinning the `src/server.ts` fallback precedence (that list also decides what
  `reconcile` *provisions*, so a wrong guess is silent until runtime).
- All 65 test projects, `lint:eslint`, `lint:types`, `api:check`, `dist:check`,
  `lint:package-json`, prettier.
- All 13 templates and 13 examples still report 0 errors unmodified. With a `SCHEDULER` binding
  injected and no export, every class-B template and class-A `examples/tanstack-start` now catch it.
  Class-A templates only miss because they ship no `lunora/_generated/` — deliberate: codegen not
  having run is not a fact about the entry.

## Reviewed

Both thermo passes ran against this diff. Findings applied, including the fail-open regression above
(reproduced end-to-end with the built CLI), a `resolveRelativeModule` bug where `./do.server` never
resolved `do.server.ts` and *did* probe a `do/index.ts` sibling, the doctor code collision, the
`workflows[]` classes being called Durable Objects, and a comment claiming five templates point `main`
at adapter output when it is one (sveltekit).

Knowingly left: the >24-module barrel budget (fail-open, and returning partial names would be
unsound), `schemaDir` not threaded from verify/doctor (pre-existing repo-wide), and doctor still
discarding the vector/global errors the FS-aware validator now computes.

Closes #652

## Summary by CodeRabbit

- **New Features**
  - Added `--env` support to the `verify` command for environment-specific configuration and binding validation.
  - Improved worker entry-point detection, including generated, composed, default-exported, and re-exported entries.
  - Scheduler support is now included in generated worker entries when configured.
  - App configuration changes now trigger a development server restart.

- **Bug Fixes**
  - `doctor` reports missing worker class exports with actionable guidance.
  - Validation warns when the configured worker entry file is missing or unreadable.
  - Additional validation findings are now surfaced instead of omitted.


packages/config/src/infer-bindings.ts
```diff
@@ -41,14 +41,63 @@ import { discoverWorkflowInfo } from "./workflow-info";
...
+const IGNORED_DIRECTORIES = new Set([".git", ".lunora-cache", ".wrangler", "dist", GENERATED_DIRECTORY, "node_modules"]);
 
 /** Directories scanned for capability signals when the caller does not override. */
 const DEFAULT_SCAN_DIRECTORIES = ["lunora", "src"] as const;
 
-/** Worker-entry candidates probed when `wrangler.main` is absent. */
-const WORKER_ENTRY_FALLBACKS = ["src/server/index.ts", "src/server/index.tsx", "src/index.ts", "src/worker.ts"] as const;
+/**
+ * Worker-entry candidates probed when `wrangler.main` names no readable source
+ * file — absent, or pointing at a framework adapter's build output.
```
prisis, line 55: Confirmed and fixed in 0d0b32b68 — thanks, this one was real and pre-existing.

Reproduced it: `main: "dist/_worker.js"` (the **built** artifact, so `existsSync` passes) with the authored entry at `src/server.ts` provisioned `[]` — not even `SHARD`. Since `reconcile` writes bindings from this, that is a green deploy that fails at runtime on a missing binding, which is exactly what the `COMPOSED_WORKER_ENTRY` docblock warns about. `src/worker.ts` only shadowed it for the templates that ship one.

The real defect was the inconsistency: `locateWorkerEntry` gained the adapter-output gate in the parent commit and this sibling did not, so the two resolvers disagreed about the same tree. `isGeneratedOutput` (plus `NON_SOURCE_DIRECTORIES` / `SOURCE_DOT_DIRECTORIES`) now lives in `infer-bindings`, which owns the entry-resolution vocabulary and is the only direction the dependency graph allows — `cloudflare/worker-entry-checks` imports from it, so it cannot import back. Kept package-internal, out of both barrels.

Pinned by a test that asserts `["SHARD"]` off the authored entry rather than `[]` off the artifact.

prisis:
Closed two of the three gaps this PR knowingly left open — `634dd931b`.

**`doctor` was dropping most of what the validator found.** It cherry-picked three findings out of the report, so an unchained `.vectors()` / `.global()`, a container image with no Dockerfile, a bad `compatibility_date` and a missing `migrations` entry all read as a clean bill of health. Remaining errors now surface as `wrangler-invalid`, remaining warnings as `wrangler-advisory` — one catch-all pair rather than a code per family, so a check added to the validator later surfaces by construction instead of being dropped again. The two already-claimed codes are excluded so `--format json` counts stay honest.

That change immediately caught this repo's own doctor fixtures declaring `ShardDO` with **no `migrations` entry** — wrangler requires one per class, so `CLEAN_WRANGLER` was never clean. Fixtures fixed, not the check.

**The barrel budget could disarm the export check silently.** `MAX_FOLLOWED_MODULES` was 24, which a barrel-heavy tree reaches, and reaching it read the entry as opaque with no signal — losing the check even for classes declared inline in the entry. It is a runaway guard, not a budget to spend (one `parseSource` per module, ~1ms with lib files skipped, against a re-export graph that is a handful of files in practice). Raised to 250, and the root no longer counts against it. Pinned by a 40-module chain that used to trip it.

**`schemaDir` is deliberately still not threaded**, and it turns out there is nothing to thread: `lunora.json` carries only `remote` and `target`, the CLI never passes `lunoraDirectory` to codegen, and `discoverSchemaInfo` is called with the literal `"lunora"`. The whole CLI assumes that directory consistently — honouring a different one in `verify` / `doctor` while codegen ignored it would be strictly worse than the assumption. Under the Vite plugin, which does know `schemaDir`, it is already passed through correctly. Making a non-default schema directory work is a new `lunora.json` field plus every command honouring it, which is its own change.

Still out of scope: class-A projects cannot use `ctx.scheduler` at all, since the composed entry cannot export `SchedulerDO`. This PR makes the check *say* so clearly instead of passing silently, but supporting it means changing what `@lunora/vite` composes — a feature, not a validator fix, and it wants its own issue.

Gates: 734 config / 1521 cli / 224 vite tests, all 65 test projects, `lint:eslint`, `lint:types`, `api:check`, `dist:check`, `lint:package-json`, prettier. 13 templates and 13 examples still clean at baseline; the P0 repro suite from the review still passes all seven cases.

PR #678, opened 2026-09-09
feat(vite,config): wire ctx.scheduler into the class-A composed entry

prisis:
> **Stacked on #677** — base is `fix/652-followup`, not `alpha`. It builds on that PR's
> `readComposedEntry`, which is what models the class-A entry's exports. Merge #677 first; this diff
> is only the feature.

## What was broken

`ctx.scheduler.runAfter` / `runAt` dispatch through a `SchedulerDO` namespace on the worker
(`create-worker.ts:133`, and the endpoints throw `SCHEDULER_NOT_CONFIGURED` without it at `:3578`).
wrangler binds only what the entry exports — and a class-A app's entry is **generated**, so there was
no file to add the re-export to. Deferred dispatch was unavailable, full stop. #677 made the validator
say so clearly instead of passing silently, which is as far as a validator can go.

Narrower than it sounds, and worth being precise: **cron jobs were never affected.** They run through
the worker's own `scheduled()` handler against the `cronJobs` map, so they work with or without this.
Only deferred dispatch and the studio's scheduled-jobs view needed it.

## The fix

The composed entry now emits:

```ts
const app = defineApp()
    .shard((env) => env.SHARD)
    .scheduler({ namespace: (env) => env.SCHEDULER })
    .httpRouter(ssrModule.default)
    .build();

export const ShardDO = app.ShardDO;

export { SchedulerDO } from "@lunora/scheduler";
```

Everything needed already existed: codegen emits the `.scheduler()` builder method when the app uses
the scheduler (`emit-app.ts:552`), and inference already tracked `usesScheduler` and even hinted at
this exact state (`infer-bindings.ts:923`). Nothing was architecturally blocked — only unwired.

## Why the binding is the opt-in, and not a plugin option

I set out to add `lunora({ scheduler: true })`. It cannot work, and the reason is worth stating
because it constrains any future version of this:

`wrangler.jsonc` is the one input `@lunora/vite`, `lunora verify`, `lunora deploy` and `lunora doctor`
**all** read. A `vite.config.ts` option is invisible to the three CLI commands — so the plugin would
export `SchedulerDO` while the export cross-check concluded it did not, and `verify` would hard-error
exactly the projects the flag enabled. Declaring the binding also keeps its `migrations` entry the
user's own deliberate act, which matters more than it looks: wrangler migration tags are append-only.

`declaresSchedulerDurableObject` is the single predicate both sides read, so the plugin's emitted
exports and the validator's model of them cannot drift. It also replaces the inline copy inside
`validateSchedulerOrigin`, so the `LUNORA_ORIGIN_URL` warning keys off the same fact — a class-A
project that declares the binding gets that reminder for free, which is the one remaining wiring step.

Two things I expected to block this and did not: the composed entry is a **virtual module**, never on
disk, so `tsc` never typechecks it (today's untyped `defineApp()` doing `env.SHARD` proves it) — so
`env.SCHEDULER` costs no typing work. And `LUNORA_ORIGIN_URL` was already validated.

## Behaviour change worth reviewing

`SchedulerDO` can no longer reach the "class-A composition cannot export this" remedy, because
declaring the binding is now its own remedy. `SessionDO` still has no route on class-A and is what
that branch now serves. The two tests that pinned the old dead-end are **retargeted onto `SessionDO`,
not deleted** — the behaviour they described is gone, but the behaviour they were protecting is not.

Auth's `SessionDO` is the same shape and one condition away if you want it; I did not build it,
because I have not checked whether auth on class-A needs anything beyond the export.

## Verification

- 736 config / 226 vite tests (4 new, 2 retargeted). The vite tests pin the coupling from the emitting
  side: the chain, the re-export, its order before `.httpRouter`, and that an app with no binding is
  byte-for-byte unchanged.
- End-to-end on `examples/tanstack-start` with the binding declared: `valid: true`, zero export
  errors, and the `LUNORA_ORIGIN_URL` warning correctly present.
- All 65 test projects, `lint:eslint`, `lint:types`, `api:check` (updated: one new exported
  predicate), `dist:check`, `lint:package-json`, prettier.
- #677's seven-case P0 repro suite still passes, and all 13 templates + 13 examples are still clean at
  baseline.

Docs: a note on the deployment page, in that file's own `:::note` convention, covering the binding as
opt-in, that crons don't need it, and the append-only migration.


prisis:
Finished the class-A story — `079e19b2a`. The scheduler was the smaller half, and my "SessionDO is one condition away" note in the description was **wrong**; here is what the code actually says.

`resolveIdentity` is set by `.auth()`, `.access()` or `.extend()` and nothing else (`emit-app.ts:868`, `:942`, `:508`) — all builder calls. The only `defineApp()` call in a class-A app lives inside the *generated* entry, so session auth was unreachable there no matter what `SessionDO` exported. `SessionDO` is a red herring: the `auth` registry item is D1-backed by default and needs no Durable Object at all.

Same seam, wider blast radius: `.global(...)` and `.vectors(...)` were also unreachable. `scanAppChains` fails open when no project file calls `defineApp`, so a class-A app declaring a `.global()` table does not fail the unchained-capability gate — it just has no global writer and throws `INTERNAL` at runtime. Verified against a scratch class-A project: `errors: 0`, and broken.

`.scheduler({ namespace: (env) => env.SCHEDULER })` was auto-composable precisely because it needs no user input. `.auth({ options: (env) => … })` takes the app’s better-auth options and cannot be derived. So what was missing is a **seam**, not an export.

**`lunora/app.config.ts` exporting `configureApp(app)`** — a conventional path, discovered rather than configured, matching `lunora/identity.ts` / `crons.ts` / `containers.ts`:

```ts
export const configureApp = (app: AppBuilder<Env>) =>
    app.auth({ d1: (env) => env.DB, options: (env) => ({ secret: env.BETTER_AUTH_SECRET }) })
       .global({ d1: (env) => env.DB });
```

The generated entry calls it between the shard selector and its own `.httpRouter(...)` / `.build()`, so framework wiring stays with the plugin. A file that exists but exports something else is ignored — the alternative is a bundle failure naming a virtual module, the least debuggable error this plugin can produce.

Two things I got wrong and fixed before pushing, both caught by checking my own output rather than the tests: the app-config import initially sat on its own template line, which left a **stray blank line in every generated entry** even for apps without the module — spurious HMR and a confusing diff on an unrelated release. It now carries its own leading newline, and a test pins the absence of churn. And the emitted specifier kept its `.ts`, unlike its sibling `_generated/app` import.

`buildWorkerEntrySource` now takes a `WorkerEntryComposition` object. Three of its parameters were adjacent booleans and this work added a seventh, so call sites read `(f, base, [], false, {}, false, path)`. That is my own mess from the previous commit, cleaned up.

**This is a new public convention**, so it is the part worth your review rather than mine. Alpha lets us rename it; after a release it is a breaking change.

Still not built: nothing further that I have evidence for. Auth on class-A now has a route, and I did not invent a second one for `SessionDO` — the DO-backed mode needs `namespace` + `internalSecret`, which belongs to whoever needs `@better-auth/scim`, not to a guess.

Gates: 736 config / 228 vite tests, all 65 test projects, `lint:eslint`, `lint:types`, `api:check` (updated for the options object + the new interface), `dist:check`, `lint:package-json`, prettier. #677's seven-case P0 suite still passes; 13 templates and 13 examples still clean at baseline.

prisis:
**Converted to draft — this PR has a boot-time break, and the docs I wrote instruct users straight into it.** Found by a thermos quality pass, confirmed against a real app in this repo.

The plugin emits `.scheduler({...})` and `export { SchedulerDO } from "@lunora/scheduler"` off the `wrangler.jsonc` binding alone. But codegen decides whether the builder even *has* that method off a completely different signal — `discover/studio-features.ts:106`: `usage.scheduler || cronCount > 0 || dependencies.has("@lunora/scheduler")`.

Verified on `examples/tanstack-start`:

```
grep -c "public scheduler(" examples/tanstack-start/lunora/_generated/app.ts   → 0
grep -c "@lunora/scheduler" examples/tanstack-start/package.json              → 0
```

So a project that declares the binding and nothing else gets `TypeError: ....scheduler is not a function` at worker boot, plus an unresolvable `@lunora/scheduler` specifier — no class-A template carries that dependency, and `lunorash` has no `./scheduler` subpath. That is the exact sequence `deployment.mdx` now prescribes: add the binding, run `lunora dev`, break. It also reintroduces a hard-coded scoped specifier into the emitted entry, which the existing regression test at `framework-compose-plugin.test.ts` was written to prevent.

My "end-to-end verification" checked the validator against that project and never the emitted entry against its own generated builder — I verified the half that agreed with me.

Second confirmed finding: `COMPOSED_ENTRY_DURABLE_OBJECTS` is still `["ShardDO"]`, so `reconcile-bindings.ts:439` now tells a correctly-wired class-A project *"@lunora/scheduler is used but the worker entry exports no SchedulerDO; export it"* — advice that is both wrong and impossible to follow, in `lunora dev` output.

The fix direction, which deletes surface rather than adding more: codegen emits `lunora/_generated/scheduler.ts` (re-exporting through the umbrella-aware specifier it already computes) when `hasScheduler`, and the plugin probes for that file exactly as it already does for `agents`/`containers`/`workflows`. Then the export, the builder call and the method availability all follow one signal by construction, the hard-coded specifier disappears, and `readComposedEntry` inherits it for free. `composesScheduler`, `SCHEDULER_CLASS` and the threaded parameter all go away.

Also to fix, both flagged and both real: `lunora/app.config.ts` collides with the root `app.config.ts` that Vinxi-era TanStack Start and SolidStart projects carry — precisely this PR's audience — and the documented `AppBuilder<Env>` example does not compile in any shipped template (no global `Env`; `@cloudflare/workers-types` provides `Cloudflare.Env`, namespaced and empty).

The correctness pass is still running. I will not push fixes until it lands, then address both sets together.

prisis:
All confirmed findings from both thermos passes are fixed — `caefd84ed`. Back out of draft.

**The blocker, at its root.** The wrangler binding was simply the wrong opt-in. Codegen decides whether the builder even *has* `.scheduler()` from `hasScheduler`, which never reads `wrangler.jsonc` — so binding-only projects composed the call onto a builder without the method and imported a package they did not depend on. Both passes landed on the same fix, and it deletes more than it adds:

Codegen now emits `_generated/scheduler.ts` off `hasScheduler`, and `scheduler` joins `GENERATED_CLASS_MODULES`. The plugin's existing per-module probe forwards it by star re-export and chains `.scheduler(...)` off that same list — so the call, the export and the method follow **one** signal by construction, and the emitted entry hard-codes no specifier. Gone: `composesScheduler`, `SCHEDULER_CLASS`, the threaded parameter, and `declaresSchedulerDurableObject`'s cross-package export. `readComposedEntry` inherits the class for free.

That also dissolves the `--env` disagreement (P0-2) rather than patching it: nothing reads `wrangler.jsonc` for this any more, so there is no top-level-vs-merged-view to get wrong.

Verified:

```
app WITHOUT a scheduler module → emits .scheduler(: false | names @lunora/scheduler: false
app WITH one                   → emits .scheduler(: true  | forwards via star: true
provisioned: ["SHARD","SCHEDULER"] | scheduler warning present: false
```

**Also fixed**

- **P1 in #677, pushed separately as `2cdf60235`** — `isGeneratedOutput` read `"./src/entry.ts"` as build output, because `.` starts with a dot. The *declared* entry was discarded, the validator blocked the deploy naming whichever fallback it probed, `@lunora/vite` threw so `lunora dev` died, and inference provisioned nothing — not even `SHARD`. Idiomatic `main`, wrangler accepts it, no template uses it, which is why CI was green. `..` had the same fault.
- **P2-6** — `inferLunoraBindings` adds `SchedulerDO` on the same condition, so reconcile provisions the binding and stops printing impossible advice on every `lunora dev`.
- **P1-4** — the `configureApp` guard parses now (`moduleExportsValue`) instead of substring-matching. The old one fired on a comment mention, a string mention, a type-only export, a default export and an unparseable file — five cases, each producing the bundle error the guard exists to prevent. All five pinned by `it.each`.
- **The name** — `lunora/app.ts`, not `app.config.ts`. You were right that it was a foreign shape: every sibling seam is a bare noun, and the dotted form collides with the **root** `app.config.ts` that Vinxi-era TanStack Start and SolidStart projects carry, which is this feature's own audience.
- **P3-8** — `lunora/app.ts` joins the dev config watch paths, so creating it mid-session no longer no-ops until a restart.
- **Docs** — the "binding is the opt-in" note was actively wrong and is rewritten (the scheduler wires itself; crons never needed it; `LUNORA_ORIGIN_URL` is the one manual step). The example gains a local `Env` and an explicit `: AppBuilder<Env>` return type — the virtual entry is in no tsconfig, so that annotation is what turns a forgotten `return` into a `lint:types` failure rather than a boot failure. Also dropped the false "verify reports both if either is missing".
- Plus `WorkerEntryComposition` exported (it is the sole parameter type of a public function), and the doc block that had drifted onto the interface.

**Not done, and I want your call rather than a guess:** the quality pass proposed going further — a codegen-emitted `defineAppConfig` marker so `Env` needs no annotation at all and discovery is AST-based like every other seam. That is the better end state, but it makes `@lunora/codegen` own the seam and I have already changed this design twice in one PR. Worth its own change.

Gates: 737 config / 236 vite / 1521 cli tests, all 65 test projects, `lint:eslint`, `lint:types`, `api:check` (3 snapshots updated), `dist:check`, `lint:package-json`, prettier. #677's seven-case P0 suite passes; 13 templates and 13 examples clean at baseline.

PR #682, opened 2026-09-10
feat(ai): let a deployment tag its AI Gateway calls

prisis:
AI Gateway can only break cost down by keys that were sent WITH the call, so a
dimension you did not tag is one you cannot recover afterwards.

Per-call tags already existed (`AiGatewayMetadata`). What was missing is the
deployment-scoped half — which app, which environment — because that is fixed at
deploy time, not at the call site. `LUNORA_AI_GATEWAY_TAGS` carries it as a flat
JSON object of string values, and every gateway call picks it up: the
`cf-aig-metadata` header for bring-your-own providers, the Workers AI binding's
native `gateway.metadata` option for the rest.

The built-in `functionPath` / `traceId` correlation fields still win over a tag
of the same name. AI Gateway rejects the whole metadata object past five keys, so
the builder trims to `AI_GATEWAY_METADATA_MAX_KEYS` rather than sending something
the gateway will throw away, and a malformed value is ignored with a warn-once —
telemetry configuration must not take inference down.

## Notes

Split out of #85 — this change has **no dependency on `apps/cloud`**; the cloud PR simply consumed it. Verified standalone on `alpha`: package tests, `lint:types`, `lint:eslint` and `api:check` all pass with only the files in this PR.

By submitting this pull request, I confirm that my contribution is made under the terms of the project's license and that you can use, modify, copy, and redistribute this contribution under those terms.

## Summary by CodeRabbit

- **New Features**
  - Added support for deployment-wide AI Gateway tags configured through an environment setting.
  - Per-request tags can be combined with deployment tags, with request-specific values taking precedence.
  - Supported metadata now includes custom tags while preserving built-in tracing information.

- **Bug Fixes**
  - AI Gateway metadata exceeding the five-key limit is now trimmed instead of rejected.
  - Invalid deployment tag configuration is safely ignored with a warning.


prisis:
**This regression cannot be caused by this PR.** The flagged benchmark is `packages/do/__bench__/count-indexed-vs-scan.bench.ts` — a Durable Object SQLite count scan. This PR changes six files, all under `packages/ai`:

```
api-snapshots/ai.api.md
packages/ai/__tests__/gateway.test.ts
packages/ai/docs/index.mdx
packages/ai/src/create-ai.ts
packages/ai/src/gateway.ts
packages/ai/src/index.ts
```

There is no path from an AI Gateway metadata builder to a D1 count query. CodSpeed flags the report itself at the top — _"Different runtime environments detected"_.

The same noise shows up on the sibling splits: #681 is flagged on `normalizeLogFields`, a pure function it does not touch, and #85 is flagged on that identical benchmark at the same ~10%.

This is the failure mode `.github/workflows/semantic-release.yml` documents in its "Refresh the CodSpeed baseline" step — release commits carry `[skip ci]`, so no benchmark run exists at `alpha`'s head and the report falls back to an older baseline, attributing everything that landed in between to the PR. `alpha` moved to `fe12556c1` shortly before this branch was cut.

Not acknowledging it on CodSpeed — that blesses a number instead of explaining it.

PR #697, opened 2026-09-11
fix(codegen): clear auth advisory on signed webhooks

prisis:
## The defect

`http_action_missing_auth_guard` fires on the canonical signed-webhook shape, and there is no way to satisfy or silence it.

The framework's own reference payment example trips it. `pnpm run codegen` in `examples/payment-demo`, **before**:

```
[WARN] http_action_missing_auth_guard — `httpAction` handler `<module>` (http:50) calls `ctx.runAction` but never reads `ctx.auth` — an anonymous caller can drive this write. Authenticate the request before the side effect.
    ↳ Read `ctx.auth` in the handler before the side effect — call `await ctx.auth.getIdentity()` (or check `ctx.auth.userId`) and reject unauthenticated/unauthorized requests, or forward through a `mutation`/`action` whose RLS policies enforce access. If the endpoint is intentionally public (e.g. a signed webhook), verify the provider signature before the write.
```

**After** — the advisory is gone, the other nine are untouched:

```
warn 9 schema advisories:
    [INFO] table_without_insert — … "customers" …
    [INFO] table_without_insert — … "events" …
    [INFO] table_without_insert — … "paymentSessions" …
    [INFO] table_without_insert — … "subscriptions" …
    [INFO] table_without_insert — … "usageEvents" …
    [INFO] procedure_without_structured_event — Public action `checkout` (billing) …
    [INFO] procedure_without_structured_event — Public action `recordApiCall` (billing) …
    [INFO] procedure_without_structured_event — Public action `apiCallsRemaining` (billing) …
    [INFO] procedure_without_structured_event — Public action `portal` (billing) …
```

The committed `examples/payment-demo/lunora/_generated/shard.ts` drops the baked-in finding to match (the only generated change in the diff).

The advisory system is not muted wholesale — `pnpm run codegen` in `examples/todo-app` still emits its three `public_mutation_without_ratelimit` WARNs plus four INFOs, unchanged.

## Why the rule was wrong

A provider webhook is authenticated by **signature**, not by identity. Stripe, Polar, Dodo Payments and Autumn send no user, so `ctx.auth` is empty on every request — reading it would be meaningless, and no edit to the handler can clear the warning. The example handler is deliberately shaped that way: it reads the raw body plus an allowlist of signature headers (`stripe-signature`, `creem-signature`, the Standard-Webhooks `webhook-*` trio, `svix-*`) and forwards them into `processWebhook`, which verifies through `ctx.payments.handleWebhook`. `httpAction` runs at the edge with the raw request *precisely so* the signature can be verified.

So every app following the documented webhook pattern got a permanent WARN it could not clear — the way a real advisory gets trained away.

## The fix, and the options rejected

**Chosen: accept evidence of signature authentication.** The codegen feeder (`packages/codegen/src/discover/http-action-guards.ts`) no longer records a handler that reads a request header whose name is signature-shaped. No new public API, no snapshot change, and it matches the house style for negative proofs in `discover/normalize-id-authorization.ts` — FN-biased suppression signals evaluated in the feeder, on the stated grounds that a false negative is cheaper than a false positive that trains users to ignore the advisor.

I checked the concern about static resolvability first, and it is real: in the example the header names go through a module-level `const SIGNATURE_HEADERS = [...]` and `request.headers.get(name)` inside a `flatMap`, so the read's own argument is a callback parameter with no literal. The check therefore resolves through the module constants the handler names — a header read anchors the claim, and the signature-shaped name may come either from the read's literal argument or from a top-level constant the handler body references. The test for that exact shape is the first of the three added below, and it fails without this resolution step.

**Rejected — a dedicated acknowledgement idiom.** There is no advisory suppression mechanism anywhere in `packages/codegen/src` or `packages/config/src` (searched `acknowledge|suppress|ignoreAdvisor|advisoryIgnore|lunora-ignore|disableAdvisor`); the house idiom is exactly what this PR does — a lint is acknowledged by writing the code that satisfies it (`public_mutation_without_ratelimit` is cleared by attaching a permissive limiter). Inventing a new opt-out for one rule would be a larger, worse change.

**Rejected — remediation text only.** The remediation already names the signed-webhook case ("If the endpoint is intentionally public (e.g. a signed webhook), verify the provider signature before the write"), and the example handler already does exactly that. Wording alone leaves the unclearable WARN in place.

## Verifying the tests by breaking the code

Three tests added to `packages/codegen/__tests__/discover/http-action-guards.test.ts`, each confirmed to fail against code without the behaviour it pins:

- *skips a signed-webhook handler that forwards a module-level signature-header allowlist* — **failed** against the unfixed feeder (expected 0 rows, got 1), passes after.
- *skips a signed-webhook handler that reads the signature header by literal name* — **failed** against the unfixed feeder (expected 0, got 1), passes after.
- *still records a handler that reads request headers carrying no signature* — the negative control. It passes both before and after by design, so it was verified by sabotage instead: forcing the signature check to `true` (suppress on any header read) makes it the only failure in the file, 1 failed / 13 passed.

## Gates

| Gate | Result |
| --- | --- |
| `pnpm run build:packages` | pass (run before the codegen re-runs, `api:check` and `lint:types`) |
| `pnpm --filter "@lunora/codegen" run test` | 139 files, 1779 tests pass |
| `pnpm --filter "@lunora/advisor" run test` | 66 files, 545 tests pass |
| `pnpm run lint:types` (repo-wide) | exit 0 |
| `pnpm run api:check` | matches all 54 committed snapshots (doc-only advisor edits; no surface change) |
| `prettier --write` then `eslint --fix` (both packages, prettier first) | clean |
| `pnpm run codegen` in `examples/payment-demo` | advisory gone |
| `pnpm run codegen` in `examples/todo-app` | still warns, unchanged |

## Deliberately left alone

The other advisories the examples trip (`table_without_insert`, `procedure_without_structured_event` in payment-demo; `public_mutation_without_ratelimit`, `nondeterministic_query_mutation` in todo-app) are a separate piece of work and are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Webhook handlers authenticated through provider signatures are no longer reported as missing `ctx.auth` checks.
  * Authentication detection now recognizes signature headers used directly or through referenced constants.
  * The same behavior applies to both HTTP actions and typed HTTP routes.

* **Documentation**
  * Updated lint guidance to clarify which webhook handlers are excluded from authentication findings.


prisis:
All three findings verified against the current code and accepted — `d15a901cb`. Two tightened the
detector (fewer false negatives), one was required to keep a legitimate shape recognised under the
tightening.

**Bind the header read to the handler request** (line 162, Security/Major) — **accepted.** Valid:
`isHeaderRead` matched any `<expr>.headers.get/has(…)`, so a handler reading a signature header off a
fetched *response* cleared the guard on its own unauthenticated write. Tightens the match. There is
now a `requestBinding` beside `contextBinding`: the second positional parameter of
`httpAction((ctx, request) => …)`, or the destructured `request` of a `.stream(…)` route. A
`.handler(…)` route receives the decoded `{ ctx, searchParams, body, params }` and no request at all,
so it can never satisfy the check — correct, since it has no inbound header to read.

**Bind signature detection to the header-read argument** (line 195, Security/Major) — **accepted.**
Valid, and the more reachable of the two: `stringLiteralsOf(body)` scanned the whole handler, so
`request.headers.get("x-tenant-id")` plus `throw new Error("missing signature")` cleared the
advisory. Tightens the match. The name now comes from the read's own argument — its string literal,
the module constant it forwards, or the constant a callback parameter iterates
(`SIGNATURE_HEADERS.flatMap((name) => request.headers.get(name))`, the reference webhook shape). An
argument that resolves to none of those yields no names, so the read proves nothing.

**Include the string-literal node itself** (line 165, Functional/Minor) — **accepted.** Valid:
`getDescendantsOfKind` excludes the node it is called on, so a single-provider
`const SIGNATURE_HEADER = "stripe-signature"` resolved to zero names and its handler kept warning.
This one *loosens* the match, so on its own it would deserve scepticism — but a constant named at the
read's argument position is the same evidence the array form already supplied, just with one
provider instead of several, and under the argument-anchored rule above it is load-bearing rather
than an extra allowance.

Each fix has a test that fails on the previous head (`6d163e0b6`):

```
 ✓ skips a named-function or wrapped handler (unresolvable body — fail-safe)
 × skips a signed-webhook handler that reads through a single-name module constant
   AssertionError: expected [ { exportName: 'webhook', …(5) } ] to have a length of +0 but got 1
 × still records a handler whose signature-shaped string is not the header it reads
   AssertionError: expected [] to have a length of 1 but got +0
 × still records a handler that reads a signature header off something other than its request
   AssertionError: expected [] to have a length of 1 but got +0

 Tests  3 failed | 14 passed (17)
```

With the fix, 17/17 pass — including the three this PR added: the module-level-allowlist webhook and
the literal `"stripe-signature"` shape still produce no advisory, and the negative control (reads
`x-tenant-id`, does a `ctx.runMutation`) still produces one.

End-to-end unchanged: `pnpm run codegen` in `examples/payment-demo` emits no
`http_action_missing_auth_guard`, and `examples/todo-app` still reports its three
`public_mutation_without_ratelimit` warnings. No generated output changed, so nothing needed
regenerating.

Gates: `build:packages` ✓ · `@lunora/codegen` 1782/1782 ✓ · `@lunora/advisor` 545/545 ✓ ·
`lint:types` ✓ (75 projects) · `api:check` ✓ · prettier ✓ · eslint ✓.

PR #713, opened 2026-09-12
security(do): re-pin caller claims across the replay gate

prisis:
## Base

Targets **`fix/r37-do-ws-client-ip`** (#702), not `alpha`. That branch reworked the same
`RequestScope` / `restoreRequestScope` pair to close the `ip` half of this bug; branching off
`alpha` would conflict on every hunk and silently revert its fix. `.coderabbit.yaml` already
lists `fix/r[0-9]+-.*` as a review base, so the stack is covered.

## The defect

`RequestScope` exists so a mutation queued at the mutation-replay gate — admitted only after a
sibling `fetch()`'s prologue has overwritten the shared `currentRequest*` fields — is restored
to its own caller context before it runs. It carried `userId` but **not the caller's claims
object**: active org, role, tenant. That is what RLS grants roles from (`readIdentityRoles` in
`@lunora/server`) and what `ctx.auth.getIdentity()` returns; the emitted `buildCtx` reads it via
`getCurrentIdentity()`.

So a queued mutation ran with its own `userId` and the sibling's role and tenant. An admin
sibling handed it admin RLS. A sibling that had already finished handed it no claims at all — a
silent downgrade. The same omission put the wrong principal on a parked or failed dispatch's
durable `__lunora_reqlog__` row and on the Logpush/SIEM event built from it.

The type's own docblock said `ip` was in the scope "for the same reason the identity fields
are", which reads as though the claims were covered. Only `userId` was. That wording is fixed
here too — it is how the gap stayed invisible.

## Reproduction

Both proofs were run against the **base branch, unfixed**, and re-run green after the change.

Node, over the real `ShardDO` dispatch path with a serializing `blockConcurrencyWhile` double:

```
AssertionError: expected { role: 'viewer', tenant: 'zinc' } to strictly equal { role: 'member', tenant: 'widgets' }

- Expected
+ Received

  {
-   "role": "member",
-   "tenant": "widgets",
+   "role": "viewer",
+   "tenant": "zinc",
  }
```

The member's `userId` was correct; the claims were the viewer's.

Real **workerd**, with only `identity` removed from the restore — a parked action's durable
request-log row picking up an interleaved sibling's claims:

```
AssertionError: expected { role: 'admin', tenant: 'acme' } to deeply equal { role: 'viewer', tenant: 'zinc' }

- Expected
+ Received

  {
-   "role": "viewer",
-   "tenant": "zinc",
+   "role": "admin",
+   "tenant": "acme",
  }
```

And the error path, where the whole emitted event is visible:

```
{"durationMs":2,"error":"boom","function":"messages:boom","identity":{"role":"viewer","tenant":"zinc"},
 "outcome":"error","source":"lunora","traceId":"6f92…","type":"request","userId":"userViewer"}
```

That is the admin's failed call, filed under the viewer.

## `currentRequest*` audit

| Field | Status | Why |
| --- | --- | --- |
| `currentRequestIdentity` | **added here** | The claims RLS reads. The defect. |
| `currentRequestTraceparent` | **added here** | `buildCtx` → `createContainerContext`; an admin `runAs` inherited a parked request's trace. Telemetry, not authorization. Also now cleared on admin entry. |
| `currentRequestBookmark` | in scope | D1 session pin; read-your-writes would read someone else's. |
| `currentRequestUserId` | in scope | The dedup namespace and `ctx.auth.userId`. |
| `currentRequestIp` | in scope | Closed by #702. |
| `currentRequestMutationId` | in scope | Replay dedup key. |
| `currentRequestClientId` / `ClientSeq` | in scope | Custom-mutator watermark identity. |
| `currentMutatorClass` | in scope | In-transaction watermark handshake. |
| `currentRequestSystem` | in scope | The `internal`-visibility gate. |
| `currentResponseBookmark` | omitted, deliberately | PRODUCED by the dispatch, not inherited — the restore clears it on purpose and the tail re-pins its own from the by-value sink. |
| `currentRequestTrace` | omitted, safe | Threaded by value as `dispatchTrace` everywhere it is read after an await, and it has its own claim/release lifecycle that a blind re-pin would fight. |
| `mutationBookkeeping` | omitted, **must stay out** | Written DURING the handler, after the capture. Re-pinning the captured `undefined` at the tail would wipe the handshake `recordPostDispatchBookkeeping` reads next. `withAdminRequestScope` keeps its own save/restore for it. |
| relation fan-out branch (`__lunora_relation__:`) | **covered (review follow-up)** | Served ahead of the ordinary path, so it awaited the child-table read before any capture existed; a throw logged the failure under a sibling. An initial scope is now captured straight after `beginDispatch`, and the later recapture is kept. |
| `currentScannedTables` / `currentIndexHits` / `currentStmtSamples` | omitted, **reported not fixed** | Per-dispatch accumulators read at the tail by `recordFunctionCall` / `flushStmtSamples`. Same interleaving class, but a sibling's prologue swaps the container object, so the writes went somewhere else — re-pinning the captured `Set` would restore the reference, not the lost stamps. Telemetry-only, and a separate change. |

## Fix shape

Added the two fields to `RequestScope` rather than replacing it with a mechanism that cannot
forget one. This is the third field found missing (bookmark, ip, now identity), so the
question was taken seriously — the argument for keeping the list is in the type's docblock:
membership is a judgement, not a category. Three per-request fields need a *different* answer
from re-pin (produced-and-cleared, threaded-by-value, written-during-the-handler), so "snapshot
everything" is actively wrong for them and any automatic mechanism would have to be told which
answer each new field wants. That relocates the judgement rather than removing it. What the
shape already buys is real: adding a field to the interface is a compile error in
`captureRequestScope`'s object literal until it is captured, so a half-made decision cannot
compile.

Two things fell out and are fixed here:

- `withAdminRequestScope` had a hand-rolled `previousIdentity` side-save precisely *because*
  the scope did not carry it; that is now redundant and gone. It also clears the inbound
  `traceparent` on entry, closing the flagged gap.
- The error path re-pins before it attributes the failure, exactly as the ok path already does
  before its own `recordRequestLog`. The capture is hoisted into a `let` the `catch` can see,
  and deliberately stays at its current statement so it still runs after `currentMutatorClass`
  is stamped — hoisting the capture itself would re-pin a blank mutator classification over
  the real one.

Not fixed, reported: the lazy `userId: () => this.getCurrentUserId()` thunks in `makeTracer` /
`instrumentDb` / `makeFetch` and `emitLog` are the same shape. They resolve at emit time, after
the handler's awaits, so they can name a sibling. Closing them means threading the caller by
value into those factories rather than extending the scope — telemetry attribution only, and a
different change.

## What the tests prove, and what they do not

- `__tests__/workerd/shard-do-request-scope.workerd.test.ts` — real workerd. An action parks on
  outbound I/O (no single-writer gate, so delivery is not blocked), a sibling `/rpc` arrives and
  runs end to end, then the parked dispatch resumes and files its request-log row. **Proves the
  interleave is reachable in production** and that the restore is what fixes it. Verified to fail
  against the unfixed restore.
- `__tests__/shard-do.request-scope-identity.test.ts` — two node-level tests against the real
  `ShardDO` over `node:sqlite`: a mutation admitted at the replay gate observing its own claims,
  and a failed dispatch's request-log row naming its own caller. The first needs a hand-rolled
  serializing `blockConcurrencyWhile`, so it settles the *code* path, not the runtime's
  scheduling.
- What none of them prove: that two concurrent `/rpc` **mutations** alone can reach the gate-queue
  window on real workerd. They cannot — while a dispatch holds `blockConcurrencyWhile`, workerd
  delivers no other events, so a second mutation's prologue never runs while the first is inside.
  A version of the workerd test built that way passed against the unfixed code and was discarded
  rather than shipped green. The gate-queue re-pin is therefore defensive for hosts whose
  `runSerialized` queues without blocking delivery (the `ShardHost` contract permits it); the
  interleave that is demonstrably live on Cloudflare today is the parked-dispatch one above.

## Review follow-up: relation fan-out (CWE-362)

A review finding said the relation fan-out branch awaits before `dispatchScope` is assigned, so a
throw from it skips the restore. **Verified and it holds.** The generated override
(`emitRelationFanout`) is `async` and awaits a whole schema-aware child-table read through
`serveRelationFanout`; the branch sits ahead of the mutator classification and so ahead of the
capture; and its throw lands in the same `catch` that calls `recordRequestLog`. Reproduced
against the previous commit — this is the emitted event for an admin's failed relation read:

```
{"durationMs":3,"error":"relation fan-out exploded","function":"__lunora_relation__:read",
 "identity":{"role":"viewer","tenant":"zinc"},"outcome":"error","source":"lunora",
 "traceId":"0e8b58c7…","type":"request","userId":"userViewer"}

AssertionError: expected 'userViewer' to be 'userAdmin'
```

Fixed as suggested: an initial `RequestScope` captured immediately after `beginDispatch`, with
the later recapture kept — the later one is the only one carrying a real `currentMutatorClass`,
so replacing it would re-pin a blank classification and break the in-transaction watermark
handshake. The `catch` guard is now unconditional.

One thing fell out on the same branch: the fan-out's **success** path echoed
`bookmarkHeaders(this.currentResponseBookmark)`, read off the shared field after the same await.
Nothing on that path ever writes that field, so the header could only ever carry a bookmark a
sibling wrote — a stranger's D1 write position reported as this caller's. It now reports the
dispatch's own by-value sink, which for a read is empty.

## Gates

| Gate | Result |
| --- | --- |
| `pnpm run build:packages` | pass (54 projects) |
| `pnpm --filter "@lunora/do" run test` | pass — 63 files, 719 tests |
| `pnpm run test:workerd` | pass — all 11 workerd suites |
| `pnpm run lint:types` | pass (75 projects; `@lunora/do` re-run uncached) |
| `pnpm run api:check` | pass — 54 snapshots unchanged, no public surface change |
| prettier, then eslint | clean |

### DIFF
--- a/packages/do/__tests__/shard-do.request-scope-identity.test.ts
+++ b/packages/do/__tests__/shard-do.request-scope-identity.test.ts
@@ -1,0 +1,247 @@
+import { AsyncLocalStorage } from "node:async_hooks";
+
+import { readRequestLog } from "@lunora/observability";
+import { runShardMigrations } from "@lunora/shard-engine";
+import { describe, expect, it } from "vitest";
+
+import type { ShardDOState } from "../src/shard-do";
+import { ShardDO } from "../src/shard-do";
+import messagesSchema from "./_helpers/messages-schema";
+import createSqliteExec from "./_helpers/node-sqlite";
+
+/**
+ * The caller-CLAIMS half of the per-request scope, over the real dispatch path.
+ *
+ * `userId` alone does not describe a caller: RLS grants roles from the claims
+ * object (`readIdentityRoles` in `@lunora/server`), and `ctx.auth.getIdentity()`
+ * returns it. A mutation queued at the replay gate is admitted long after a
+ * sibling `fetch()`'s prologue has overwritten the shared `currentRequest*`
+ * fields, so anything the gate does not re-pin is the sibling's — which for the
+ * claims means the queued mutation evaluates RLS as whoever ran while it waited.
+ */
+
+/** One caller's identity headers, kept next to the claims they encode. */
+interface Caller {
+    claims: Record<string, unknown>;
+    userId: string;
+}
+
+const ADMIN: Caller = { claims: { role: "admin", tenant: "acme" }, userId: "userAdmin" };
+const MEMBER: Caller = { claims: { role: "member", tenant: "widgets" }, userId: "userMember" };
+const VIEWER: Caller = { claims: { role: "viewer", tenant: "zinc" }, userId: "userViewer" };
+
+/** What the handler saw on `this` at the moment it ran. */
+interface Observation {
+    functionPath: string;
+    identity: Record<string, unknown> | undefined;
+    userId: string | undefined;
+}
+
+/**
+ * A shard whose handler records the caller context the generated `buildCtx`
+ * reads (`getCurrentUserId` / `getCurrentIdentity`), and which can park a named
+ * function until the test releases it.
+ *
+ * `messages:send` and `messages:update` are mutations — the kind that takes the
+ * `runSerialized` replay gate; `messages:poll` is an action, which does not.
+ */
+class ScopeObservingShard extends ShardDO {
+    public readonly observed: Observation[] = [];
+
+    public readonly parks = new Map<string, Promise<void>>();
+
+    public override async handleRpc(functionPath: string): Promise<unknown> {
+        const park = this.parks.get(functionPath);
+
+        if (park !== undefined) {
+            await park;
+        }
+
+        this.observed.push({ functionPath, identity: this.getCurrentIdentity(), userId: this.getCurrentUserId() });
+
+        if (functionPath.endsWith(":boom")) {
+            throw new Error("boom");
+        }
+
+        return { ran: functionPath };
+    }
+
+    /** Park `functionPath` inside the handler until the returned callback runs. */
+    public park(functionPath: string): () => void {
+        let release: () => void = () => {};
+
+        this.parks.set(
+            functionPath,
+            new Promise<void>((resolve) => {
+                release = resolve;
+            }),
+        );
+
+        return release;
+    }
+
+    // eslint-disable-next-line class-methods-use-this -- pure predicate over the path, mirroring the codegen override
+    protected override isMutationFunction(functionPath: string): boolean {
+        return functionPath === "messages:send" || functionPath === "messages:update";
+    }
+}
+
+const makeState = (database: ReturnType<typeof createSqliteExec>): ShardDOState => {
+    return {
+        acceptWebSocket() {},
+        getWebSockets() {
+            return [];
+        },
+        storage: { sql: database.sql as unknown as ShardDOState["storage"]["sql"] },
+    };
+};
+
+/**
+ * A `blockConcurrencyWhile` double that actually serializes, so the second
+ * mutation really queues instead of interleaving. `AsyncLocalStorage` marks the
+ * callback's own await chain so a NESTED `runSerialized` (what
+ * `runInTransaction` does) runs in place rather than deadlocking on the gate its
+ * own caller holds — workerd's behaviour. Mirrors the double in
+ * `shard-do.idempotency.test.ts`.
+ */
+const makeSerializedState = (database: ReturnType<typeof createSqliteExec>): ShardDOState => {
+    let queue: Promise<void> = Promise.resolve();
+    const insideGate = new AsyncLocalStorage<true>();
+
+    return {
+        ...makeState(database),
+        blockConcurrencyWhile: async <T>(callback: () => Promise<T>): Promise<T> => {
+            if (insideGate.getStore() === true) {
+                return callback();
+            }
+
+            const previous = queue;
+            let release: () => void = () => {};
+
+            queue = new Promise<void>((resolve) => {
+                release = resolve;
+            });
+
+            await previous;
+
+            try {
+                return await insideGate.run(true, callback);
+            } finally {
+                release();
+            }
+        },
+    };
+};
+
+const rpcRequest = (functionPath: string, mutationId: string, caller: Caller): Request =>
+    new Request("https://shard.internal/rpc", {
+        body: JSON.stringify({ args: {}, functionPath }),
+        headers: {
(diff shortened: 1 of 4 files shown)


prisis:
Confirmed and fixed in afc509e70.

**Verification.** The finding holds on all three points. `runRelationFanoutRead`'s generated override (`emitRelationFanout` in `packages/codegen/src/emit.ts`) is `async` and awaits a full schema-aware child-table read via `serveRelationFanout`; the branch is served ahead of the mutator classification and so ahead of the capture; and a throw from it lands in the same `catch` that calls `recordRequestLog`. Reproduced against the previous commit with a sibling parked on the shared fields — the emitted event for an admin's failed relation read:

```
{"durationMs":3,"error":"relation fan-out exploded","function":"__lunora_relation__:read",
 "identity":{"role":"viewer","tenant":"zinc"},"outcome":"error","source":"lunora",
 "traceId":"0e8b58c7…","type":"request","userId":"userViewer"}

AssertionError: expected 'userViewer' to be 'userAdmin'
```

**Fix**, in the suggested shape: an initial `RequestScope` captured immediately after `beginDispatch`, and the later recapture kept. Two captures rather than one moved capture, for the reason you gave — the later one is the only one carrying a real `currentMutatorClass`, and hoisting it would re-pin a blank classification over the real one and break the in-transaction watermark handshake. The `catch` guard is now unconditional.

**One thing fell out on the same branch.** The fan-out's *success* path returned `bookmarkHeaders(this.currentResponseBookmark)`, read off the shared field after the same await. Nothing on that path ever writes that field, so the header could only ever have carried a bookmark a sibling wrote during the await — a stranger's D1 write position reported as this caller's. It now reports the dispatch's own by-value sink, which for a read is empty. Noted as a behaviour change in the commit body: a successful relation read no longer carries a `bookmark` header.

Regression test: `files a FAILED relation fan-out read under the caller that made it`, in `packages/do/__tests__/shard-do.request-scope-identity.test.ts`. Verified failing against the previous commit (output above) before the fix landed.

Audit table in the PR body updated with a row for the relation fan-out path.

Gates re-run on the new head: `build:packages` pass · `@lunora/do` 63 files / 719 tests pass · `test:workerd` all 11 suites pass · `lint:types` pass · `api:check` 54 snapshots unchanged · prettier then eslint clean.

PR #719, opened 2026-09-12
fix(platform-node): emulate workflow step retries and rollbacks

prisis:
The Node workflow host accepted two documented `WorkflowStepLike` features and used neither:
`createStepAdapter` hardcoded `attempt: 1` (so `config.retries` was ignored) and took the rollback
argument as an unused `_rollback?: unknown` (so a step's `rollback` never ran). The capability
matrix declared neither gap.

## Reproduction

A workflow with a `rollback`-bearing step followed by a `retries: { limit: 3 }` step that always
throws, run against the real host with `MemoryStore`:

```
status=errored flakyRuns=1 rollbacks=0
```

After the fix, same workflow, same harness:

```
status=errored flakyRuns=3 rollbacks=1
```

## What I chose, and why

**Emulate both, rather than declare-only.** The task offered "declare the gap in
`PlatformCapabilities` so codegen can diagnose it" as the cheap option, and I checked whether that
was actually reachable before picking. It is not: `gateAgainstMatrix` gates on the whole `workflows`
key, and there is no finer-grained key for per-step retries or compensation. Rating `workflows`
`unsupported` to declare a partial gap would gate every workflow off the Node host — a far bigger
regression than the defect. Adding a `workflowStepRetries`-style key would be what
`capabilities.ts`'s own header calls "half a change": codegen cannot see per-step config inside a
handler body, so the key would ship as an advisory rating that warns nobody.

That leaves the note text as the only honest carrier at this granularity, and a note gates nothing.
Since declaring could not actually protect anyone, emulating was the option that closes the hole.
Both emulations fell out of the one `ctx.step` call the adapter already makes:

- **Retries** — an attempt loop inside the memoized step, with constant/linear/exponential backoff
  and a real 1-based `ctx.attempt`. `limit` is read as the total number of attempts, which is what
  this file's previous `{ retries: { limit: 1 } }` default already meant by it.
- **Rollbacks** — completed steps with a handler are collected per activation and unwound in
  reverse declaration order when a step finally fails, the failing step's own handler first with
  `output: undefined`. The engine replays the workflow body from the top on every activation, so a
  memoized step re-registers its rollback before a later step can fail — that is what makes the
  unwind whole across a suspension, and it has its own test.

**`config.timeout` is deliberately left un-emulated** and is now declared as a gap. The contract
hands the callback no `AbortSignal`, so this seam could only reject while the work it was meant to
cancel kept running — a step that reports failure and then completes its side effect is worse than
one that takes too long.

## Capability ratings, before and after

No rating level changed. `workflows` was and remains `emulated` — the honest reading both before
and after, since the engine does the durability and this host does the rest. What changed is the
note, which is the matrix's only carrier at this granularity and is held verbatim by
`pnpm run lint:node-capabilities-docs`.

| feature | before | after | note change |
| --- | --- | --- | --- |
| `workflows` | `emulated` | `emulated` (unchanged) | Now states that retries and rollbacks are the adapter's rather than the engine's, with their crash boundaries (a crash mid-backoff restarts the step at attempt 1; a crash mid-unwind leaves compensation half-done; `rollbackConfig` is ignored), and lists `config.timeout` as a gap. |
| `websocketHibernation` | `emulated` | `emulated` (unchanged) | See the audit below. |

The source-of-truth note in `packages/platform/src/capabilities.ts` and the hand-written copy in
`packages/platform-node/docs/index.mdx` were updated together from one script, so the two are
byte-identical and `lint:node-capabilities-docs` passes.

## Audit of the host's other ratings

I checked every other `NODE_CAPABILITIES` rating against the implementing source rather than
against its own note. One more overstatement, corrected here:

**`websocketHibernation`** claimed "attachments/tags persisted to SQLite, **so subscription state
survives a process restart**". The rows do persist and `createNodeSocketHost` does expose a
`restoreSocket` that rehydrates one — but `restoreSocket` is not a member of the `SocketHost`
contract, and both composition roots throw it away:
`node-platform.ts:211` destructures only `socket`, and `node-shard-registry.ts:262` takes
`createNodeSocketHost(database).socket`. Its only non-test caller in the repo is the conformance
host. So nothing composed by `createNodePlatform` or `createNodeShardRegistry` reassociates a
reconnecting client. The `emulated` rating is still right — the durable half is real — but the
claimed consequence is not, and the note now scopes itself to what the code reaches.

Checked and found accurate (implementation read, each specific claim in the note verified):
`localSql`, `shardedState`, `shardAlarms`, `scheduler`, `cronTriggers`, `queues`, `objectStorage`,
`objectStorageCdcArchive`, `objectStorageBackups`, `keyValueStore`, `globalTables`, `memoryTables`,
`crossShardFanout`, and the `unsupported` set (`mail`, `agents`, `vectorStore`, `secrets`,
`httpCache`, `identityProxy`, `durableStreams`).

Two I did **not** clear: `serverReactors` and `commitOrderedTables`. Both notes attribute the
behaviour to `@lunora/shard-engine` plus this host's write gate. The write gate checks out
(`withBoundary`, `node-shard-host.ts:547-605`), but I did not read the engine's reactor and
commit-sequence code, so I am not calling them verified.

One wording nit left alone as not worth a churned note: `shardAlarms` says "6 attempts, from 100ms",
which is 6 re-deliveries on top of the initial one.

## Tests

Five new cases in `packages/platform-node/__tests__/node-workflow-host.test.ts`, each run against
both stores (`MemoryStore` and `createNodeWorkflowStore`) by the file's existing `describe.each`:
retry-then-succeed with the attempt counter asserted, give-up-after-the-last-attempt, reverse-order
unwind with outputs, compensation of a step that completed **before a suspension** (the replay
case), and a throwing rollback neither masking the step error nor stranding the rest of the unwind.

Verified by breaking the code first, not just by passing:

- Pinning `const limit = 1` (the old behaviour): `4 failed | 38 passed`, with
  `expected 'errored' to be 'complete'` and `expected 1 to be 2`.
- Removing the `await compensate(...)` call: `6 failed | 36 passed`, with
  `expected [] to strictly equal [ { output: undefined, …(1) }, …(2) ]` and
  `expected [] to strictly equal [ 'reserve' ]`.

## Gates

| gate | result |
| --- | --- |
| `pnpm run build:packages` | pass — 54 projects |
| `@lunora/platform-node` tests | pass — 180 tests, 15 files |
| `@lunora/platform` tests | pass — 51 passed, 1 skipped |
| `@lunora/workflow` tests | pass — 123 tests |
| `@lunora/codegen` tests | pass — 1776 tests, 139 files |
| `pnpm run lint:node-capabilities-docs` | pass |
| `pnpm run lint:types` | pass — 75 projects; also re-run uncached as a bare `tsc --noEmit` in each of the four touched packages |
| `pnpm run api:check` | pass — "Public API surface matches all 54 committed snapshots" |
| `prettier --check` (4 changed files) | pass |
| `eslint` (`@lunora/platform-node`, `@lunora/platform`) | pass |

`test:workerd` was not run: this change touches no SQL, storage or Durable Object path — the
workflow host is pure Node over the `@visulima/workflow` engine.

One note on `lint:types`: its first run failed with `TS2307: Cannot find module
'@lunora/shard-engine'` in that package's own `__bench__` file. It did not reproduce on a re-run,
and it does not reproduce on a clean tree with these changes stashed either — a dist race in the
first build, not this diff.

## Summary by CodeRabbit

- **New Features**
  - Workflow steps now support automatic retries with constant, linear, or exponential backoff.
  - Completed steps can register rollback actions that run in reverse order when a step ultimately fails.

- **Known Limitations**
  - Step timeout settings are not enforced by the Node adapter.
  - Retry and rollback behavior is emulated within the adapter; an interruption during backoff may restart from the first attempt.


prisis:
All three verified against the branch before touching anything; all three were real. Fixed in 14f37008.

### 1 — compensation order recorded at completion, not declaration · **accepted**

Confirmed. `ctx.parallel` spawns branches through `Promise.all` of `deps.step.do(...)` (`fan-out.ts`), and `ctx.runStep` forwards rollbacks straight to `step.do` (`run-step.ts`), so a `Promise.all` of `runStep` reaches the same shared adapter. A step declared first but settling last was pushed last, and `toReversed()` then unwound it first.

New test, against the unfixed branch:

```
FAIL  __tests__/node-workflow-host.test.ts > createNodeWorkflowHost — 'MemoryStore' > unwinds concurrent steps in declaration order, not completion order
FAIL  __tests__/node-workflow-host.test.ts > createNodeWorkflowHost — 'createNodeWorkflowStore' > unwinds concurrent steps in declaration order, not completion order
AssertionError: expected [ 'ship', 'slow', 'fast' ] to strictly equal [ 'ship', 'fast', 'slow' ]

- Expected
+ Received

  [
    "ship",
-   "fast",
    "slow",
+   "fast",
  ]
```

`slow` (declared first, 20 ms) and `fast` (declared second, immediate) run under one `Promise.all`, then a third step fails. Fixed as proposed: `Compensation` gains an `order` assigned from a per-activation counter read at the top of `do`, before the call awaits anything, and `compensate` sorts by it descending instead of reversing arrival order. The counter is rebuilt on replay alongside the compensation list, so it stays consistent across a suspension. Sequential bodies are unchanged — the existing five cases all still pass.

### 2 — non-finite retry limits · **accepted**

Confirmed, with a bounded probe (callback stops throwing after 1000 attempts, so a honoured limit caps well below that):

```
limit 3        -> 3 attempts
limit NaN      -> 1000 attempts
limit Infinity -> 1000 attempts
```

`NaN`/`Infinity` never satisfy `attempt >= limit`; the probe only terminated because the callback gave in.

Kept at the adapter rather than moving it to `defineStep`, for the reason you flagged: the adapter is where every path converges. A direct `step.do(name, config, cb)` never passes through `defineStep`, and `ctx.runStep(step, args, { config })` overrides whatever a definition declared — so a `defineStep` guard would cover neither. `VALIDATION_ERROR`, consistent with the `toMs` non-finite-duration guard already in the file. Two cases added (`Infinity`, `NaN`); both bodies succeed on their first attempt on purpose, so a regression reports a completed run the assertions reject rather than spinning the loop until CI is killed.

### 3 — oversized timer clamping · **accepted** (finding 2's guard does *not* make it unreachable)

I checked the interaction you asked about, and it goes the other way. `"1 month"` parses to 2 592 000 000 ms — already past the 2 147 483 647 ceiling — so a plain `retries: { delay: "1 month", limit: 2 }` reaches it with no exponential growth and a perfectly finite limit. Observed by spying on `setTimeout` during a real run:

```
setTimeout delays seen: [ 2592000000 ] | MAX: 2147483647
```

Above the ceiling Node takes the delay as 1 ms, so the backoff meant to spare a failing dependency became back-to-back retries. `delay` now clamps to `MAX_TIMER_MS`. That is a ceiling, not a repair: the backoff is still an in-process `setTimeout` that would outlive the activation's lease long before 24.8 days elapse. Clamping only stops an oversized backoff from becoming no backoff, and the comment says so.

### Capabilities note

Unchanged, and still correct — the `workflows` note already described the unwind as reverse declaration order and listed `config.timeout` as the gap. This change makes the first claim true rather than aspirational; it adds no `ctx.*` surface and shifts no rating, so `docs/index.mdx` needed no edit.

### Gates

`build:packages` · `@lunora/platform-node` 188/188 · `@lunora/platform` 51 passed, 1 skipped · `@lunora/workflow` 123/123 · `lint:node-capabilities-docs` · `lint:types` (`--no-cache` on the three projects) · `api:check` (54 snapshots, after a fresh build) · prettier · eslint — all green, prettier before eslint, and re-checked after ESLint's autofix so the two are stable together.

Every new test was verified to fail against the unfixed code before the fix landed.

PR #736, opened 2026-09-12
fix: four defects found by building a SaaS kit on the packages

prisis:
## Summary

Four defects in `@lunora/cli`, `@lunora/react` and the `payment` registry item. Each surfaced from a gate rather than a review while composing the shipped items into a real multi-tenant app, and each hits any project doing the same thing.

**`lunora registry add` scaffolded a schema codegen cannot read.** On a project with no `lunora/schema.ts`, it wrote `export const schema = defineSchema({})` — but the emitted `app.ts` and `shard.ts` import that module's **default**, so every such project failed `tsc` with `TS2613` until someone changed the export by hand. `insertSchemaExtension` already resolves the chain from the `defineSchema(...)` call rather than a variable binding, so it splices into either shape; only the stub was wrong.

**`payment`'s `mySubscriptions` projected three fields, none of which a billing screen can use.** No `priceId`, so nothing matches a subscription to a plan; no `cancelAtPeriodEnd`, so nothing can say "cancels at period end" — which outranks the state, since a subscription can be `active` *and* ending. Widened to what a screen actually reads, with `quantity` marked as seats **billed**, because that number lags an invite by however long a webhook takes and should not be counted for display.

**`APP_BASE_URL` reached the running Worker and not the type-checker.** The item scaffolded it into `.dev.vars` only, and binding types are generated from wrangler's config — so reading the item's own env var was a `TS7053`. It is now declared as a `vars` binding as well. The other half is a project prerequisite rather than an item bug, and the item's docs now name it: the types only apply when a project's tsconfig `include` covers wrangler's generated `worker-configuration.d.ts`.

**`useMutation` and `useAction` return handles that read like functions.** `const send = useMutation(api.x.y); send({ … })` is a `TS2349`. Both docblocks now say so and show the destructuring (`mutate` and `call` respectively).

The binding-kind check reported `vars` as an unknown kind, which was it working — a new kind needs declaring. It is the one kind that is a name→value map rather than a list of resource entries, so it has no wrangler-required fields, and empty is not the same as absent.

### Deliberately not fixed here

A fifth finding — **`createAuth` erases better-auth's plugin inference** — is real and is *not* in this PR. `createAuth(options: LunoraAuthOptions): LunoraAuth` is annotated with the non-generic instance type, so enabling `organization()` or `admin()` gives you `session.activeOrganizationId` and `user.role` at runtime and nothing at the type level. Two fixes were tried and both fail for principled reasons worth recording:

- Making `createAuth` generic works for consumers, but `Auth<O>` is **invariant** in `O`, so every helper annotated with the erased type then rejects a plugin-carrying instance. Making those helpers generic in turn breaks their own bodies: `auth.api.*` cannot resolve against an unbound type parameter.
- Widening the alias to `Auth<any>` restores assignability, but the `any` propagates into the helper bodies and erases the type information they rely on (21 new errors in `admin.ts` alone).

The direction that does work is retyping each helper by the narrow structural surface it uses rather than by the whole instance — a package-wide refactor of `@lunora/auth`, which is stable-tier and deserves its own review rather than riding along here.

## Test plan

- [x] `pnpm --filter @lunora/cli run test` — 1582 passing, one pre-existing `init.test.ts` failure that reproduces on a clean tree
- [x] `pnpm --filter @lunora/react run test` — 202 passing
- [x] `pnpm --filter @lunora/cli run lint:eslint`, `@lunora/react` likewise — clean
- [x] `pnpm run lint:registry:sync` and `lunora registry build --check` — clean
- [x] End-to-end for the `vars` fix: composed `payment` into a real project, ran `wrangler types`, `tsc --noEmit` clean once the generated types are in `include`
- [ ] Reviewers: re-run `pnpm test` on a fresh build — several CLI failures on this branch turned out to be stale `dist/`, not code

## Checklist

- [ ] Updated relevant docs in `apps/docs` — package-local docstrings and the item's `docs` field were updated instead
- [x] No new package added
- [ ] No migration impact

## Notes for reviewers

Start with `registry/payment/payment.ts`: the widened projection is the change most likely to affect someone already reading `mySubscriptions`, though it only adds fields.

The `vars` entry in `REQUIRED_BINDING_FIELDS` is a test-fixture change, not a loosened assertion — the check now knows the kind rather than reporting it as unrecognised.

https://claude.ai/code/session_01UaKquhzCKTDQsBiuqwXmrV

---
_Generated by [Claude Code](https://claude.ai/code/session_01UaKquhzCKTDQsBiuqwXmrV)_

---

## Review round (thermo passes)

Five commits on top, after rebasing onto the base branch. Two reviewers and a
local gate sweep converged on one high-severity finding on the money path.

### The `vars` entry shipped a localhost checkout URL to production

`vars` is *deployed* configuration and `.dev.vars` wins locally, so
`APP_BASE_URL: "http://localhost:5173"` was read in exactly one place:
production. That turned a loud failure into a silent one — before, the var was
absent in production and `appOrigin()` threw on the first `checkout`; after,
`checkout` returned 200 and handed Stripe a `success_url` on the paying
customer's own machine. The entry now ships **empty**: `wrangler types` still
emits `APP_BASE_URL: string`, so the `TS7053` this was added to fix stays
fixed, and the existing `value === ""` guard still throws. That also resolved
the manifest disagreeing with itself (`envVars` said `:8787`, the binding said
`:5173`).

**Root cause, fixed separately:** `lunora deploy` has a gate for exactly this,
and it filtered a hardcoded list of *variable names* — so it covered
`LUNORA_ORIGIN_URL` and `AUTH_URL` and nothing else. It now filters on the
value being a loopback URL, which is where the invariant actually lives, and
covers `APP_BASE_URL` and `PUBLIC_STORAGE_BASE_URL` too.

### `priceIds` had drifted across four mirrors

Entitlements test membership in `priceIds`, not equality with `priceId` — a
Stripe subscription is a list of items. The canonical column says so: *"Apps
that mirror these tables inline must add the column to get multi-item
entitlements."* Four mirrors hadn't:

- `packages/react/src/payment.tsx` — **its drift gate is red on the base
  branch**, so `pnpm --filter @lunora/react test` failed before this.
- `registry/payment/schema.ts` — claims to mirror the canonical tables and
  didn't, so every scaffolded project got single-item entitlements.
- `examples/payment-demo` — what an author copies the inline declarations from.
- `mySubscriptions` — widened here, but documenting `priceId` as the plan key.

The projection also regained `provider` (the discriminator that lets two
providers coexist mid-migration) and `currentPeriodStart` (without it a client
falls back to `createdAt` and shows *lifetime* usage against a per-period
limit), and `state` is the `SubscriptionState` union rather than `string`. The
`row["x"] as string` casts became narrowing reads — an assertion types a
missing column as `string` while handing the client `undefined`.

### Placement and coverage

- The hook "not callable" note covered 2 of **4** such hooks, cross-referencing
  each other so it read as exhaustive; `useMutator` (`mutate`) and `useAgent`
  (`run`) were missing. The canonical version now lives in the Convex migration
  guide — where `useMutation(...)` *is* the invoker, so this is a migration trap
  first — and the docblocks are one line each.
- The schema-stub explanation moved out of the middle of a ternary into a named
  `schemaStub`, absorbing a single-use local.
- `vars: []` in `REQUIRED_BINDING_FIELDS` gave that map a second meaning;
  modelled as `MAP_SHAPED_BINDINGS` instead, so the "unknown binding kind"
  branch keeps its teeth. `vars` also joined the object-vs-array shape check.
- The manifest `docs` string (1510 chars, printed raw as one terminal line) gave
  its `TS7053`/tsconfig paragraph back to the README's env-var table, which is
  the prose that survives into a consumer's repo. The README also listed three
  things `registry add payment` does when it now does four.

### Verified

`api:check` · `lint:generated` (all 17 generators) · `lint:registry:sync` ·
`lint:package-json` · `lint:prettier` · `dist:check` (on a production build) ·
`registry build --check` · eslint + `lint:types` for `cli`/`react`/`payment`/
`payment-demo` · **all 11 `test:workerd` suites** · `@lunora/cli` 1585,
`@lunora/react` 204, `@lunora/payment` 296 — all passing.

The three red checks on the previous head were all the stale base: the
worker-size gate's JSON-envelope fix had already landed upstream. The
`init.test.ts` failure the test plan called pre-existing is also gone.

New tests were each confirmed to fail against the unfixed code.

## Summary by CodeRabbit

- **New Features**
  - Subscription data now supports multiple price or product IDs for multi-item billing.
  - Subscription details include richer cancellation, billing, pricing, provider, quantity, and status information.
  - Payment setup declares `APP_BASE_URL` for generated environment types and supports named Wrangler environments.

- **Bug Fixes**
  - Deploy checks the full IPv4 loopback range when validating local URLs.
  - Missing schema files now receive a compatible default schema.

- **Documentation**
  - Clarified payment configuration requirements and hook invocation patterns.


packages/cli/src/commands/deploy/handler.ts
```diff
@@ -1253,7 +1246,17 @@ const checkLocalhostOriginVariables = (cwd: string, logger: Logger, command: Pre
...
 
-    const offenders = ORIGIN_VAR_NAMES.filter((name) => typeof variables[name] === "string" && isLocalhostUrl(variables[name]));
+    // Every `var` whose value is a loopback URL, not a list of known names: the
+    // invariant is a property of the VALUE (a deployed Worker cannot reach
+    // loopback, whatever the var is called), and the allowlist this replaces had
+    // already gone stale — it covered `LUNORA_ORIGIN_URL` and `AUTH_URL` while
+    // `APP_BASE_URL` and `PUBLIC_STORAGE_BASE_URL` shipped localhost defaults
+    // past it. A name list has to be extended by whoever adds the next origin
+    // var, which is exactly the person who doesn't know this gate exists.
+    const offenders = Object.entries(variables)
```
prisis, line 1268: Leaving this chain as it is.

The list is a project's `vars` block from `wrangler.jsonc` — a handful of keys, walked once in a deploy preflight that is about to spawn wrangler. The two extra passes are not measurable next to that, and the rule's target (a chain re-run on every render) does not apply to a one-shot CLI check.

The declarative form is also what makes the invariant readable: filter by value, project to names, sort for a stable message. Folding it into a `reduce` or a `for...of` with a mutable accumulator would hide that behind loop bookkeeping to save nothing.

---

prisis:
**CodSpeed Performance Analysis** is red on the current head, and the regression is not this change's. Recording why, since nothing here will make it go away.

The one benchmark that moved is `in-batch: single IN(...) query + id->doc re-projection` in `packages/do/__bench__/rank-page-hydrate.bench.ts` (789 µs → 948.2 µs), with 287 untouched. That file is self-contained: it imports one `node:sqlite` test helper, creates its own table, inserts its own 1000 rows, and times statement execution. Nothing this PR changes is imported by it, directly or transitively — the diff is CLI command code, React docblocks, the payment registry item, and docs.

The stronger evidence is that the same benchmark moved by the same amount (-16.12%) on this PR's first head, whose entire React change was comment text. Comments cannot move a Durable Object benchmark. Two heads with disjoint content producing the same single-benchmark delta, against two different bases, is runner variance — and the earlier run said as much, warning that the comparison spanned different runtime environments. The `n+1` benchmark alongside it in the same file, on the same SQLite handle, did not move.

So there is no fix to port and nothing to widen this PR with. The check can be cleared by acknowledging the regression in CodSpeed, which needs a maintainer.

For completeness on the other two reds earlier in this PR's history: `Worker size budget` was genuine and is fixed here by merging the base branch, which carries the change that reads the bundle measurement from the JSON envelope — it is green now. The `Check Lint Run` / `Check Test Run` failures before that were aggregator jobs reporting a run that a new push had cancelled mid-flight.

This benchmark job's one re-run is spent: its first attempt failed before any benchmark body ran (`Installation of CodSpeed CLI with version 4.17.6 failed. Reason: curl: (35) Recv failure: Connection reset by peer`), and the re-run is what produced the measurement above.

---

packages/cli/src/commands/deploy/handler.ts
prisis, line 1268: Same finding on the same line as the earlier thread, re-posted against the new head. The chain stays as it is, for the reasons recorded there: a handful of `wrangler.jsonc` keys walked once in a deploy preflight, where the declarative filter/project/sort is what makes the invariant readable.

---

packages/cli/src/commands/deploy/handler.ts
```diff
@@ -1253,7 +1246,17 @@ const checkLocalhostOriginVariables = (cwd: string, logger: Logger, command: Pre
...
+        .filter(([, value]) => typeof value === "string" && isLocalhostUrl(value))
```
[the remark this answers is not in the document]
prisis: Good catch — real, and fixed. `http://127.0.0.2:8787` did parse to hostname `127.0.0.2`, fail the equality check, and ship an origin the Worker cannot reach.

Fixed with a dotted-quad match rather than the proposed prefix, because `hostname.startsWith("127.")` also matches `127.example.com` — a legal, routable DNS name — and would block a real deploy. The gate now tests `/^127(?:\.\d{1,3}){3}$/`, which covers the whole `127.0.0.0/8` block and nothing else.

The shorter numeric spellings needed no pattern of their own: `new URL()` canonicalises them for http(s), so `127.1`, `0x7f.1` and `2130706433` all arrive at the check already spelled `127.0.0.1`.

The test asserts both halves — `127.0.0.2` blocked, the normalising shorthand still blocked, `127.example.com` not blocked. Against the old check it fails exactly the way the bug reads: the shorthand is caught and `127.0.0.2` walks through.

---

prisis:
A third CodSpeed run settles the question I raised above, so recording it rather than leaving the earlier comment as the last word.

On this head the flagged set changed completely:

| Run | Head | Flagged |
| --- | --- | --- |
| 1 | first head | `rank-page-hydrate` (do) −16.12% |
| 2 | after the base merge | `rank-page-hydrate` (do) −16.79% |
| 3 | current | `v.string()` (values) −14.59%, `import-throughput` (runtime) −10.36% |

`rank-page-hydrate` — the benchmark my earlier comment was about, and the one CodSpeed twice called a regression — is now among the 286 untouched, with no change to it in between. The two benchmarks now flagged live in `packages/values` and `packages/runtime`.

The only code change between run 2 and run 3 is the loopback fix: one hoisted regex and one test, both in `packages/cli`. The PR touches no file under `packages/values`, `packages/runtime` or `packages/do` at all.

A set of "regressions" that rotates between runs while the measured code never changes is the runner, not the diff. Same conclusion as before, now with the earlier candidate exonerating itself. Still nothing to port and nothing to widen this PR with; clearing the check needs the regressions acknowledged in CodSpeed, which is a maintainer action.

---

prisis:
Fourth CodSpeed run, and two things in it close this out.

`rank-page-hydrate` — the benchmark the first two runs called a **−16% regression** — is now reported as a **+20.52% improvement** (949.8 µs → 788.1 µs). Nothing in this PR has ever touched it. A measurement that swings −16% then +20% on the same untouched code is measuring the runner.

The report also now names its own limitation: *"No successful run was found on `alpha` (ae74b6b) during the generation of this report, so b4a3a95 was used instead as the comparison base. There might be some changes unrelated to this pull request in this report."* So the comparison is not against this PR's base, and CodSpeed says as much.

The one flagged regression this time is `normalize-log-fields` in `packages/do`, again untouched by the diff. Position is unchanged: nothing to fix, nothing to port, and the check clears by acknowledging in CodSpeed.

---

PR #756, opened 2026-09-13
fix(sql-store): guard the global unique-index and cdc-rewind paths

prisis:
Two defects on the `.global()` (D1) plane of `packages/sql-store`. Each has a
shard-plane sibling already in `alpha`, and the interesting part of both is where
the D1 answer is *not* the DO answer.

## S1 — a UNIQUE index over rows that already violate it

Both producers of a UNIQUE index on a global table — a declared `unique: true`
index and a `.unique()` column — reached `CREATE UNIQUE INDEX` unguarded. The
duplicate probe the drop side already had (`dropIndexIfShapeChanged`) never
covered the create side.

**What actually happens on D1**, measured rather than assumed:

```
THROW#1: Error | UNIQUE constraint failed: users.email | code=ERR_SQLITE_ERROR
THROW#2: UNIQUE constraint failed: users.email
```

A bare engine `Error`. Not a `LunoraError`, no table named, no remedy. And it is
not confined to the offending table — a read of an unrelated table on the same
store fails with `no such table`, because the provisioning loop bails at the
first throw and never reaches the tables declared after it:

```
THROW: UNIQUE constraint failed: aaa.slug
tables: [ 'aaa' ]          ← "zzz" was never created
```

So the blast radius is **wider** than the shard plane's (whole global plane, not
one shard), and the recovery is **easier**: `ensureMigrated` deletes its cache
entry on rejection, so each request retries cleanly rather than latching, and D1
is reachable out-of-band (`wrangler d1 execute`, the migration runner, studio) to
de-duplicate — where a wedged DO cannot even run the de-dup `defineMigration`,
since `runShardDataMigration` goes through `ensureMigrated()` first. It is a
total outage of the global plane, not an unrecoverable one.

### Where this diverges from the shard-plane fix

| | shard plane (`shard-engine`) | here |
|---|---|---|
| when the probe runs | before the create, gated on a `sqlite_master` "is the index held" read | **after** the create has failed, gated on `dialect.isUniqueViolation(error)` |
| engines covered | SQLite | SQLite, Postgres, MySQL |
| cost on the happy path | one catalogue read per unique index per pass | none |

The probe-first shape does not transfer well. A shard runs its migration pass
once per wake; a host builds one ctx-db **per request**, so a catalogue read per
unique index per unmigratable-or-not provisioning is a per-request cost the DO
never pays. Probing only after a real failure costs nothing when there is
nothing wrong, and needs no per-engine "does this index exist" query — which is
why the guard reaches Postgres and MySQL, where `dropIndexIfShapeChanged` is
SQLite-only.

Gating on `isUniqueViolation` is what makes it safe: the original error always
propagates, so an unrelated DDL failure is never relabelled — not even on a table
that *does* hold duplicates, which a probe run unconditionally would have blamed.
There is a test for exactly that.

**Not ported, because it is already shared:** the drift-gate half. `diffIndexes`
lives in `shared/schema-snapshot.ts` and already classifies an added unique index
as `severity: "breaking", remediation: "backfill"` for both planes.

## S2 — a rewound `.global()` changelog

`readSqlCdcChanges` answered any `sinceSeq` past the end of the log with an empty
page and the cursor echoed back. Reproduced:

```
cursor after 3 writes: 3
(restore)
post-restore writes land at seq 1, 2
page from sinceSeq=3:  0 changes, cursor 3   ← "you're caught up"
rows actually in log:  seq 1, seq 2          ← both unreachable, forever
```

### What a D1 restore reverts, and what survives

Time Travel restores the **whole database**. Nothing inside it survives — which
is the same conclusion the shard plane reached about a DO's native restore, and
the reason an epoch row can never report the event it exists for.

What differs is which witnesses exist outside. The shard plane has two (a
consumer's cached cursor, and the R2 changelog archive). **The global plane has
one.** There is no `.global()` changelog archive — `cdcArchiveRewound` and
`ctx-db-cdc-archive.ts` are shard-only. So the consumer's own cursor is the whole
of the evidence, and the guard acts on it where it is presented.

The cursor-above-watermark witness does work here, with one correction to the
watermark. `MAX(seq)` is wrong: a retention sweep DELETEs rows, so a log swept
empty reports NULL and every healthy consumer would be accused of holding a
rewound cursor. `sqlite_sequence` — the AUTOINCREMENT bookkeeping row — survives
the DELETE and rewinds only with the database around it. Verified against real
D1 under workerd, not just `node:sqlite`:

```
sweptMaxSeq: null      sweptSequence: 2      afterSweep: served
                                             afterRestore: threw:CDC_TIMELINE_FORKED
```

SQLite only, the way `dropIndexIfShapeChanged` in this package already is.
Postgres and MySQL have a sequence too, but reading it is dialect-specific
catalogue SQL against engines this package has no suite for, and MySQL's
`information_schema` AUTO_INCREMENT is a cached estimate — a wrong watermark here
refuses a consumer that is perfectly in sync.

### The part I did not build, and why

**An epoch on the global changelog is a design decision, not a port.** It needs
saying plainly because the shard-plane fix has one and this does not.

The watermark guard closes the window between the restore and the moment
post-restore writes climb back past a consumer's cursor. After that the witness
is gone, and on the shard plane the re-minted epoch is what still refuses a
consumer reconnecting later. Reproducing that here needs three things, and the
third is the blocker:

1. a durable epoch row on the global plane — new table, new migration. Cheap.
2. re-minting it when a witness proves the fork. Cheap.
3. **a channel that carries the epoch to a consumer and back.** There isn't one.
   `readSqlCdcChanges` / `readD1CdcChanges` are bare exported helpers that users
   wire into their own sync endpoints; there is no subscription frame and no
   `cdcSync` admin RPC on this plane. Adding one means a new field on the return
   value, a new input, and every external warehouse connector persisting and
   echoing it — a protocol change to a published API whose consumers are
   user-written.

Options, with their costs:

- **(a) Epoch returned on every page, consumer echoes it.** Closes the window
  fully. Breaks every existing connector until it is updated, and silently
  degrades to the current behaviour for any that is not, since an absent echoed
  epoch is indistinguishable from a consumer that never had one.
- **(b) Epoch folded into the cursor** (`"<epoch>:<seq>"`). No new parameter, and
  an old numeric cursor stays parseable. But the cursor is currently a `number`
  in the return type and in the connector contract, so it is still a wire break,
  just a quieter and therefore worse-behaved one.
- **(c) Watermark guard only** — what this PR does. No API break, no consumer
  change, and it catches the case that actually loses data in practice: a restore
  under a live connector, which notices on its next poll rather than after the
  log has caught back up.

(c) is what holds on its own evidence. (a) vs (b) is a call about the public CDC
contract that wants an owner, and I am not making it inside a bug fix.

**Also deliberately not changed:** the `.global()` shape poll
(`ShardDO.globalPollCursor`). It reads `readSqlCdcChangedTables`, not this path,
and its cursor is an in-memory field — a rewind freezes shapes for at most
`GLOBAL_SHAPE_RESYNC_MS`, after which the unconditional `cursorOnly` pass adopts
the lower `MAX(seq)` and it self-heals. Bounded and self-correcting; no guard
earns its place there.

## API

`cdcForkedError` gains a `scope`, as `cdcTrimmedError` already has, and its
`epoch` becomes optional — the global refusal has none to carry. One call site
(`ShardDO.runShardCdcSync`) updated; `api-snapshots/shard-engine.api.md` moves by
one line. No new error code: `CDC_TIMELINE_FORKED` is already in the catalog, so
the errors snapshot and the generated error reference are untouched.

## Verification

`build:packages`, `api:check`, prettier, eslint, `vis run lint:types --no-cache`,
`--filter @lunora/sql-store test` (214), `--filter @lunora/d1 test` (331),
`--filter @lunora/do test` (749), `--filter @lunora/shard-engine test` (1499),
and `test:workerd` — all 11 suites — pass.

Each guard was verified by breaking it after the fix was committed and confirming
the intended test went red: the guard removed, each producer routed around it,
the `isUniqueViolation` gate dropped, the NULL filter dropped, the boundary
flipped to `>=`, the watermark read from `MAX(seq)`, the empty-log fallback
changed, and the two probes reordered. Two of those go red **only** under
workerd, against real D1.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved change-feed error reporting when clients request data beyond the available history, distinguishing global and shard-level timelines.
  - Added detection for rolled-back database timelines to prevent consumers from reading invalid change sequences.
  - Improved unique-index creation and migration checks, including clearer handling when duplicate records prevent an index from being created.


prisis:
`Lint (eslint)` (and the `Check Lint Run` aggregate it feeds) is red here for a reason that is not in this diff.

The failure is `examples/payment-demo/lunora/billing.ts:106:31 — arrow-body-style`. That file is byte-identical to `origin/alpha` (`git diff origin/alpha HEAD -- examples/payment-demo/lunora/billing.ts` is empty) and this branch touches nothing under `examples/` at all. `alpha` is red on that check, so every open PR inherits it; #758 fixes it.

Everything else this branch is responsible for is green, including all 11 `Workerd integration` jobs (`d1` and `do` among them), `Lint (types)`, `Lint (api surface)`, `Lint (prettier)`, `Lint (dist production)`, `Lint (generated files)`, E2E, and `MySQL integration (@lunora/hyperdrive)`.

PR #761, opened 2026-09-13
feat(saas-ui,templates): add the composed SaaS kit

prisis:
## Summary

A composed SaaS starter, built from the packages this repo already ships rather than beside them. The finding that started it (`plans/454-saas-kit.md`): Lunora owned every ingredient — auth with organisations, payments across six providers, ratelimit, storage, mail, presence — and composed none of them into a runnable multi-tenant app, so every evaluator had to assemble one to see whether the framework worked.

Three layers, each usable without the ones above it:

**`registry/saas`** — the shared backend. `saas_projects` and `saas_activity` shard by `organizationId`; `saas_organizations` is `.global()`. One copy, framework-independent.

**`packages/saas-ui`** — the kit's logic with no framework import: 11 core modules driving the dashboard, tables, billing, org management and presence. `vitest.config.ts` compiles `core/` with no plugin, which is the property that enforces the boundary. Distributed as `saas-ui-react` and `saas-ui-svelte`, gated by `lint:registry:sync`.

**`templates/saas`** — the kit as a running app (TanStack Start + React). The hand-authored frame is 12 files; the `lunora/` payload beneath it was composed by the real `registry add`, which pulled `saas-ui-react` → `saas` → `auth` + `ratelimit` in one command. `lunora/server.ts` carries the two callbacks that make tenancy real: `resolveIdentity`, and an `authorizeShard` admitting a caller to their own organisation's shard and the root shard, nothing else.

Svelte exists as the proof view deliberately — no JSX, compiler reactivity, a render model maximally unlike React's. A port that agrees with React about everything proves nothing. **Neither port changed a line of `core/`**, which was the plan's STOP condition.

### The merge, and what it surfaced

The branch had fallen ~850 commits behind. The merge was textually clean — the kit almost entirely adds files — but one semantic conflict was underneath it, fixed here:

`templates/saas/lunora/payment/schema.ts` inlines the payment tables, making it a **third copy** of the canonical schema, and it was missing `priceIds`. Entitlements test membership in that list, so a subscription billing a base plan alongside an add-on has a `priceId` naming only one of them — every project started from this template would have had single-item entitlements. That is exactly what the canonical column's docstring warns inline mirrors about. The drift gate named one copy; it now takes a list, so a copy added later is one array entry from coverage rather than a silent fork.

### What the rest of the gates found

Running the gates that do not fit in `lint` and `test` turned up three more defects in the template, each of which would otherwise have shipped to every project scaffolded from it:

- **The `.global()` organizations table had no writer.** `saas_organizations` serves the two reads a shard cannot — the admin's cross-organisation lists, and resolving an organisation before its shard is known — but nothing chained `.global({ d1: … })` onto `defineApp()`. A scaffolded project compiled cleanly and then threw `INTERNAL` on every admin page and every slug lookup. The registry item had the same hole in its install instructions rather than its code; both say it now.
- **`APP_BASE_URL` was committed as `http://localhost:5173` in `wrangler.jsonc`.** That key is deployed configuration, so the value is read only in production, where it is wrong — and the guard meant to catch it could never fire: `checkout` would succeed and hand the payment provider a `success_url` on the customer's own machine. The `payment` item already ships it empty for exactly this reason and documents why; the template now agrees, with the dev value in `.dev.vars`.
- **The template's `saas-ui` payload had forked from the registry item.** It is a verbatim copy written at compose time and never re-run, so it kept a `useForm` that mutated a ref during render, a form controller without `resetOnSuccess`, and the barrel imports whose cycle caused a TDZ crash — all three already fixed upstream. Refreshed, and gated by a byte-comparison mirror test beside the payment one.

Also, three error codes the kit minted (`FAILED_PRECONDITION`, `INVALID_ARGUMENT`, `ALREADY_EXISTS`) are not in `ERROR_CATALOG`, so the wire would have redacted each to a generic 500 and the friendly message keyed on it could never appear. They are now `UNPROCESSABLE` / `VALIDATION_ERROR` / `CONFLICT`, the vocabulary every other registry item speaks. The rate-limit mapping had the same shape of bug: it keyed on `RATE_LIMITED`, which only a token budget throws, while the `rateLimit` middleware throws `TOO_MANY_REQUESTS`.

## Test plan

- [x] `pnpm --filter @lunora/saas-ui run test` — 95 passing across 10 files, 4 vitest projects
- [x] `pnpm --filter @lunora/saas-ui run lint:types` — `tsc --noEmit` plus `svelte-check`, 317 files, 0 errors
- [x] `pnpm run lint:registry:sync` — both families up to date
- [x] `pnpm --filter @lunora/cli` mirror gates — payment (11) and saas-ui (25); each verified to FAIL on exactly the drifted file when a copy is edited
- [x] `pnpm run test` over the workspace — green
- [x] `test:templates` — the saas leg passes scaffold, install, `lunora add auth-ui`, build, typecheck (with its coverage floor) and the credential-free deploy dry run
- [x] `api:check`, `dist:check`, `lint:project-json`, `lint:catalog-drift`, `lint:generated`, `lint:peer-catalog-refs`, `lint:package-json`, Prettier, ESLint — all green
- [x] `test:workerd` (11 packages), E2E, the Worker size budget and the benchmark suite — green in CI against the new base

## Checklist

- [ ] Updated relevant docs in `apps/docs` — the kit documents itself in `plans/454-saas-kit.md`, the template README and the registry items' `docs`; the docs-site page is workstream L, deliberately not in this PR
- [x] New package added: `packages/saas-ui` has its `project.json` tags, its `.agents/docs/packages.md` row and an api-snapshot tier
- [ ] No migration impact

## Notes for reviewers

Start with `packages/saas-ui/src/core/` and one view beside it — that pair is the architectural claim, and it is the thing a fifth and sixth view port will either confirm or break.

Then `templates/saas/lunora/server.ts`, which is where tenancy is enforced rather than described. It carries one documented cast: `createAuth` erases better-auth's plugin inference, so `session.activeOrganizationId` exists at runtime and not at the type level. That defect is real and is not fixed here — two approaches were tried and both fail for principled reasons; the direction that works is retyping each `@lunora/auth` helper by its narrow structural surface, which is a package-wide refactor of a stable-tier package and deserves its own review.

CodeRabbit skipped this PR: 112 files against its 100-file limit. Splitting it would mean drawing boundaries the change does not have — the registry item, the package and the template are one composed artifact, and the gates that caught the defects above are precisely the ones that compare them against each other.

Known gaps, all tracked as workstreams in the plan rather than hidden: the app shell (sidebar, ⌘K, URL-synced table state, RBAC-filtered nav), app admin, four remaining view ports, an e2e run of sign up → create org → invite → accept, and the kit's own bundle budget entry.

https://claude.ai/code/session_01UaKquhzCKTDQsBiuqwXmrV


prisis:
**CodSpeed Performance Analysis** is the one red check on `946a582` (53 of 54 others are green). It is not this PR's regression, and there is no fix to port — the measurements moved, the code did not.

Two benchmarks are reported slower:

- `packages/observability/__bench__/metrics-emit.bench.ts` — `count, 3 attributes → SQLite select + upsert + prune`, 304.2 µs → 409.4 µs
- `packages/do/__bench__/trigger-overhead.bench.ts` — `1 after-insert no-op trigger`, 292.9 µs → 331.5 µs

Why it is not this diff:

1. **This PR changes no file in either package.** `git diff --name-only <base>...HEAD` touches `templates/saas`, `packages/saas-ui`, `registry/saas*`, `packages/cli`, plus the expo template/example and a few scripts. `packages/observability`, `packages/do` and `packages/shard-engine` — the three packages owning all four benchmarks that moved — have **zero** changed files. Nothing in the new code is reachable from these benchmarks.
2. **The job that produced the numbers passed.** The `Benchmarks` workflow job concluded `success`; only CodSpeed's own threshold verdict on the uploaded samples is red.
3. **CodSpeed itself flags the comparison as unreliable.** Its report carries "⚠️ Different runtime environments detected — Some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results", plus an unexpected-base footnote.
4. **The movement is symmetric noise.** 2 regressed, 2 improved, 284 untouched — and the two improvements (`ReactiveCache.run` miss +36.05%, `normalizeLogFields` +13.3%) are in code this PR does not touch either. A real regression from this diff would not come with matching speedups in unrelated packages.

I am not acknowledging the regressions on CodSpeed and not touching benchmark thresholds — neither would be mine to decide, and both would hide a signal that belongs to whoever owns those packages. Re-running the benchmark workflow once to re-measure under a fresh runner; if it comes back red with the same two entries in untouched packages, the verdict is an environment artifact on this branch's baseline rather than anything this PR introduced.

---

